PR #5, opened 2016-03-07
Switches to using public docker hub

gxb5443:
Also updates to go1.6 and makes push script a little more dynamic.


poopoothegorilla:
+1

PR #6, opened 2016-03-09
Python plugin

gxb5443:
Adds support for compiling protobufs to Python.  Also adds enhancements to build script.


Dockerfile
```diff
@@ -17,21 +17,22 @@ RUN set -ex \
...
 	&& make install \
   && cd \
-	&& rm -rf /mnt/protobufs/ \
-	&& apk --no-cache add libstdc++ \
+	&& rm -rf /tmp/protobufs/ \
+  && rm -rf /tmp/protobufs.tar.gz \
+	&& apk --no-cache add libstdc++ \ 
 	&& apk del .pb-build \
-  && rm -rf /usr/local/include \
+  #&& rm -rf /usr/local/include \
```
poopoothegorilla: do we want to just remove this? or write why it is commented?

python/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+MAINTAINER Core Services <coreservices@namely.com>
+
+ENV GRPC_RELEASE release-0_13
+
+# Install protoc python generator
+###################################
+
+RUN set -ex \
+	&& apk --no-cache --update add --virtual .python-builder \
+  make \
```
poopoothegorilla: marked

poopoothegorilla:
+1

Dockerfile
bobbytables: I concur.

PR #18, opened 2016-06-08
Bump golang to build an image.

bobbytables:


karlkfi:
Did you get an image published? looks like dockerhub still says 1 month ago.

poopoothegorilla:
@karlkfi should be pushed now

PR #20, opened 2016-08-01
Compile csharp proto without grpc plugin

johnnymo87:
grpc is not _yet_ available for dotnet core, so [our dotnet core app](https://github.com/namely/Namely.Payroll.Data) needs to use a [protobuf-only protoc](https://github.com/google/protobuf) instead of a [grpc-friendly protoc](https://github.com/grpc/grpc/tree/master/src/csharp).

TODO:
- [x] Build a docker image that exposes a protobuf-only protoc
- [ ] Push image to our registry

To better explain the difference, here's the git diff of some code I'm generating from a proto file I have.

```
+        errorMessage_ = pb::Preconditions.CheckNotNull(value, "value");
-        errorMessage_ = pb::ProtoPreconditions.CheckNotNull(value, "value");

```

The `-` part of this is the generated code with the grpc mixins, the `+` part is without it.

### DIFF
--- a/dotnet-core/Dockerfile
+++ b/dotnet-core/Dockerfile
@@ -1,0 +1,5 @@
+FROM namely/protoc
+MAINTAINER Core Services <coreservices@namely.com>
+
+COPY script.sh /opt/namely/script.sh
+ENTRYPOINT ["/opt/namely/script.sh"]
--- a/dotnet-core/script.sh
+++ b/dotnet-core/script.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+TARGET_DIR="pb-csharp"
+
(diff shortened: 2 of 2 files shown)


therealplato:
i see no obvious problems, but then i don't know what to look for

johnnymo87:
OK, I'm ready to move forward with this, and I think my first step is to commit and tag my change to the image as [described here](https://docs.docker.com/engine/tutorials/dockerimages/). I was looking for how we're currently doing this in [our docker registry](https://registry.namely.run/repository/), but I don't see any `namely/protoc-*` repos there. Do I not have permission to see them, or are they really not there?

johnnymo87:
Closing in favor of https://github.com/namely/docker-protoc/pull/23

PR #26, opened 2017-11-16
create protoc-all container

mhamrah:
This creates a new namely/protoc-all container supporting all languages and plugins. It will deprecate all other container usage. During the cutover period the other containers will remain as is, untouched.


README.md
```diff
@@ -25,22 +22,11 @@ So if you have a directory: `/Users/me/project/protobufs/` that has:
 
 ```sh
 cd ~/my_project/protobufs
-docker run -v `pwd`:/defs namely/protoc-ruby
+docker run -v `pwd`:/defs namely/protoc-all -f myproto.proto -l ruby #or go, csharp, etc
```
johnnymo87: I've had a hell of a time getting volumes to work like this in bash for windows. `docker-compose`, however, works like a charm.

mhamrah: Interesting, with volume mounts? It's probably worthwhile adding a docker-for-windows example too.

mhamrah:
Quickly spoke to Abe seems like powershell is much better for
Docker-on-windows and volume mounts.

mhamrah:
This has been pushed to dockerhub for those wanting to try it out: `docker.io/namely/protoc-all`

all/entrypoint.sh
```diff
@@ -0,0 +1,117 @@
...
+
+if [ $GEN_GATEWAY = true ]; then
+    mkdir -p ${GEN_DIR}/pb-go
+    protoc $PROTO_INCLUDE \
+		--grpc-gateway_out=logtostderr=true:${GEN_DIR}/pb-go \
+        $FILE
+	protoc $PROTO_INCLUDE  \
+		--swagger_out=logtostderr=true:${GEN_DIR}/pb-go \
+        $FILE 
+fi
```
bobbytables: Rad. 👍

mhamrah:
yeah I was thinking that too- I've been manually doing it with the
test.proto file. I can add something that runs the script for supported
languages and checks for exit 0 and/or the files.

test.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Generate proto files
+for lang in ${LANGS[@]}; do
+    echo "Testing language $lang"
+    docker run --rm -v=`pwd`:/defs namely/test-protoc-all -f test.proto -l $lang
+    if [[ ! -d gen/pb-$lang ]]; then
+        echo "generated directory does not exist"
+        exit 1
+    fi
+done
```
bobbytables: 👍

PR #30, opened 2017-12-04
Fixes to grpc-gateway generation

mdkess:
grpc-gateway generation is now golang specific.

gateway files are generated in the same directory as the protobufs, which is required for imports to work.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -8,7 +8,10 @@ docker build -t namely/test-protoc-all ./all
 for lang in ${LANGS[@]}; do
     echo "Testing language $lang"
     docker run --rm -v=`pwd`:/defs namely/test-protoc-all -f test/test.proto -l $lang -i test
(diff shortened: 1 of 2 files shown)


test.sh
```diff
@@ -8,7 +8,10 @@ docker build -t namely/test-protoc-all ./all
 for lang in ${LANGS[@]}; do
     echo "Testing language $lang"
     docker run --rm -v=`pwd`:/defs namely/test-protoc-all -f test/test.proto -l $lang -i test
-    docker run --rm -v=`pwd`:/defs namely/test-protoc-all -d test -l $lang -o gen/dir/$lang --with-gateway
+    if [[ "$lang" == "go" ]]; then
```
mhamrah: nit: remove go check and move outside of the for loop. this will test the command with and without the `--with-gateway` flag for go

mdkess:
how's this?

mhamrah:
awesome 🥇

PR #31, opened 2017-12-05
script to generate docker container for grpc-gateway.

mdkess:
Example usage:

`bash generate_gateway.sh -f FunTimes.proto -s FunTimes -c mkess-test-gateway`

Then run the container with
`docker run -p 8080:80 mkess-test-gateway --backend=mkess-cool-backend-service`


all/generate_gateway.sh
```diff
@@ -0,0 +1,300 @@
...
+HTTP_PORT=8080
+
+printUsage() {
+  echo "Generates a docker image of a grpc-gateway server for a given service proto."  
+  echo
+  echo "The container will listen on port ${HTTP_PORT} for HTTP traffic, and proxy requests"
+  echo "to the gRPC service. In addition, it will serve a Swagger definition of its API at"
+  echo "/swagger.json"
+  echo
+  echo "Usage: generate_gateway.sh -f <service.proto> -p service_package -s ServiceName"
```
mhamrah: -p is service_package here, put below it's port.

mhamrah: we need to support multiple proto files. this bit me. it is important that all proto files are passed to `protoc` at once. `entrypoint.sh` was recently changed to better handle this.

all/generate_gateway.sh
```diff
@@ -0,0 +1,300 @@
...
+    exit 1
+fi
+
+if [[ -z $CONTAINER ]]; then
+    echo "Error: You must specify the Docker container name"
+    printUsage
+    exit 1
+fi
+
+if [[ -z $GRPC_SERVICE_HOST ]]; then
```
mhamrah: I don't think this and GRPC_SERVICE_PORT should be set here. I think the must be configured through env variables, or via flags on the container's entrypoint. https://github.com/spf13/viper has a good configuration hierarchy, and the main.go method should ensure these are set and present.

all/generate_gateway.sh
```diff
@@ -0,0 +1,300 @@
...
+
+  SignalRunner(
+    func() {
+      fmt.Printf("launching http server on %v\n", server.Addr)
+      if err := server.ListenAndServe(); err != nil {
+        log.Fatalf("Could not start http server: %v", err)
+      }
+    },
+    func() {
+      shutdown, _ := context.WithTimeout(ctx, 10*time.Second)
```
bobbytables: Is there a reason you're not getting the CancelFunc from this and defer'ing it being called?

bobbytables:
Make sure you update the README too to change the "ideally it would include a container blah blah" at the end.

all/generate_gateway.sh
mdkess: Fixed yo.

mdkess:
I updated this - it generates the image for you. Right now it has to run on the client's machine, next step is to run it in the container (i.e. using /defs or whatever to mount the dir).

all/generate_gateway.sh
mhamrah: There's no use for the cancel func in any scope, is there?

mhamrah:
+1. my comment is more of a nit, but i think it will be beneficial.

also, we can probably forego putting this in a container?

mhamrah:
Can this be merged?

PR #32, opened 2017-12-07
Added grpc-gateway swagger protos to dockerfile.

mdkess:


all/Dockerfile
```diff
@@ -21,6 +21,7 @@ COPY install-protobuf.sh /tmp
 RUN chmod +x /tmp/install-protobuf.sh
 RUN /tmp/install-protobuf.sh
 RUN git clone https://github.com/googleapis/googleapis
+RUN git clone https://github.com/grpc-ecosystem/grpc-gateway
```
mhamrah: this should already be present in the final container because we `go get` it. You can copy the files into the include directory from `/go/src/github.com/grpc-ecosystem/grpc-gateway` to the include dir with a vanilla `RUN cp` I believe

all/entrypoint.sh
```diff
@@ -136,6 +136,7 @@ esac
 
 PROTO_INCLUDE="-I /usr/include/ \
     -I /usr/local/include/ \
+    -I /usr/include/grpc-gateway/ \
```
mhamrah: you should be able to just make this `-I/go/src/github.com/grpc-ecosystem/grpc-gateway`

mdkess:
Done

mhamrah:
Did you push?

PR #38, opened 2018-01-25
Added grpc_cli dockerfile

mdkess:


README.md
```diff
@@ -54,15 +54,49 @@ $ docker run ... namely/protoc-all -f protorepo/catalog/catalog.proto -l go
...
+alias grpc_cli='docker run -v `pwd`:/defs --rm -it namely/grpc-cli'
+```
+
+Note the use of single quotes in the alias, which avoids expanding the `pwd` parameter when the alias
+is created.
+
+Now you can call it with
+
+```sh
+grpc_cli call docker.for.mac.localhost:50051 LinkShortener.ResolveShortLink "short_link:'asdf'" --protofiles=link_shortener.proto
```
mhamrah: wortwhile adding a multi-file example to avoid users googling.

cli/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# Grab standard proto libraries
+WORKDIR /tmp
+RUN git clone https://github.com/google/protobuf.git
+RUN git clone https://github.com/googleapis/googleapis
+RUN git clone https://github.com/grpc-ecosystem/grpc-gateway
+
+
+# Build release image
+FROM debian
```
mhamrah: protoc-all uses alpine. Any reason for the switch?

cli/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+WORKDIR /tmp/grpc
+RUN git submodule update --init
+
+# Build grpc_cli
+# For some reason, we have to do this manually.
+WORKDIR /tmp/grpc/third_party/protobuf
+RUN ./autogen.sh
+
+WORKDIR /tmp/grpc
+RUN make grpc_cli
```
mhamrah: Why not run this in https://github.com/namely/docker-protoc/blob/master/all/Dockerfile and just create a new dockerfile with protoc-all as base and a new entrypoint? You'd remove this file and duplication completely.

cli/Dockerfile
mdkess: Yeah - I couldn't find libgflags-dev for Alpine, which is required by grpc. That's also the answer to the other question.

README.md
mdkess: Good call! Done!

cli/Dockerfile
mhamrah: You can grab it from a mirror, see https://github.com/tombh/texttop/blob/master/Dockerfile.

mhamrah: may be better: https://gist.github.com/dukelion/2a8813f8adb4ac08b500175f4daa88fd

mdkess:
As discussed - I'll submit this as is, but I'll try to go back and consolidate stuff later.

PR #39, opened 2018-01-30
Made python use underscores instead of dashes

mdkess:
Also added __init__.py files so that generated module can be imported.

This is because python cannot (easily) import from directories with
a dash in them.


mhamrah:
retroactive 👍

PR #44, opened 2018-03-05
Added working php plugin

ayenter:
Now works with php language using php plugin.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ test.pb.go
 test.rb
 gen/*
 pb-*
+.idea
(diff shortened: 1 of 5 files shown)


bobbytables:
Hi @ayenter - Sorry we took a while to look at this PR. Would you mind rebasing against master? I'd love to get the doc work in this PR in. I'd like to test the PHP generation as well.

mhamrah:
PHP was added in #47, wonder if this is still needed?

mhamrah:
closing due to inactivity, and php is supported

PR #47, opened 2018-03-21
Add gogo and php grpc support

codyaray:


mhamrah:
Thanks for this! Curious how are you finding gogo support compared to vanilla go?

mhamrah:
Would you mind adding an entry to `all/test.sh`?

codyaray:
Sure! We're using gogo almost exclusively nowadays.

PR #49, opened 2018-03-21
Add support to generate nodejs output.

sundbp:


sundbp:
Not producing solid output yet. Will revisit with a new PR later.

PR #51, opened 2018-03-22
Add support to generate nodejs output.

sundbp:
The node grpc plugin currently has a hard coded assumption that the message file
to go with a certain service file is "../foo/foo_pb.js". This is not always the
case, but for now we try to ensure the --js_out argument is set to match what
the grpc plugin expects.

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -144,7 +144,6 @@ case $GEN_LANG in
         GEN_STRING="--grpc_out=$OUT_DIR --${GEN_LANG}_out=$OUT_DIR --plugin=protoc-gen-grpc=`which protoc-gen-grpc-java`"
         ;;
     "node")
-        SERVICE_NAME=$(basename $PROTO_DIR)
         GEN_STRING="--grpc_out=$OUT_DIR --js_out=import_style=commonjs,binary:$OUT_DIR --plugin=protoc-gen-grpc=`which grpc_${PLUGIN_LANG}_plugin`"
         ;;
     *)
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -19,7 +19,7 @@ printUsage() {
 }
 
 GEN_GATEWAY=false
-SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc")
+SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc" "node")
 EXTRA_INCLUDES=""
 OUT_DIR=""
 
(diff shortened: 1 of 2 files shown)


all/entrypoint.sh
```diff
@@ -143,6 +143,10 @@ case $GEN_LANG in
     "java")
         GEN_STRING="--grpc_out=$OUT_DIR --${GEN_LANG}_out=$OUT_DIR --plugin=protoc-gen-grpc=`which protoc-gen-grpc-java`"
         ;;
+    "node")
+        SERVICE_NAME=$(basename $PROTO_DIR)
```
mdkess: I don't think you need this `SERVICE_NAME` line.

all/entrypoint.sh
```diff
@@ -143,6 +143,10 @@ case $GEN_LANG in
...
+        GEN_STRING="--grpc_out=$OUT_DIR --js_out=import_style=commonjs,binary:$OUT_DIR --plugin=protoc-gen-grpc=`which grpc_${PLUGIN_LANG}_plugin`"
```
mdkess: Question for you - how useful are these extra items (i.e. `import_style=commonjs,binary`)? Do you think that other people in the community might want different ones?

all/entrypoint.sh
```diff
@@ -19,7 +19,7 @@ printUsage() {
 }
 
 GEN_GATEWAY=false
-SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc")
+SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc" "node")
```
mdkess: What do you think about calling this `js` instead of `node`? I think it'd at least have consistency with the `$LANG_out`, and might be useful for other javascript frameworks.

Can you also update all/test.sh to include this in the list of languages to test? Line 3 in that file should be all that you have to change.

all/entrypoint.sh
sundbp: indeed - it's a leftover. I'll remove.

all/entrypoint.sh
sundbp: I can only get commonjs style to output consistent files when using gRPC. The binary flag seems "standard". It's possible that people may want other combinations, but these ones seem pretty consistent with what happens for other languages in the image.

all/entrypoint.sh
sundbp: The protoc pluging is called node, that's why I chose it. I think it's nodejs only as it stands. Can change it to js - then need to change the plugin part to not use the LANG but instead use a fixed js. Swings and roundabouts on that one.

I'll add the test.

sundbp:
Found this project a long while back through a blog post from namely btw :)

mhamrah:
Stick with node. It differentiaties between browser compatible code and a
node just runtime-specifically the c layer.

mdkess:
I built and pushed a new `namely/protoc-all` container image.

Thanks again for the contribution! We really appreciate it 👍

mhamrah:
Did you use the 1.9 tag?

sundbp:
Thanks! Much appreciated - I've found it a sane way to develop with protobuf and gRPC.

PR #52, opened 2018-03-24
bump to 1.10

mhamrah:


pawmart:
all good but we need to update docker hub instance too ;/

PR #57, opened 2018-04-03
Allow running gateway under directory.

### LINKED ISSUE #56: Allow configurable prefix for the API endpoints
Imagine if we wanted to host all of our APIs under the /api endpoint.

A service's gRPC gateway may expose an endpoint such as `/my-service/foo`. Clients would access it by going to `/api/my-service/foo`. A rewrite rule at the edge proxy (i.e. NGINX) would redirect requests to `/api/my-service` to the grpc-gateway.

However, with the current implementation of grpc-gateway, the grpc-gateway wouldn't understand the path.

This task is to make the grpc-gateway aware, via a configuration value, of the prefix under which it's running. That is, `/my-service/foo` defined in the proto would be accessible via `/api/my-service/foo`.

mdkess:
By specifying proxy.api-prefix, you can have the gateway strip prefixes.

For example, suppose that you have a proto file with the following:

```
option (google.api.http) = {
  get: "/v1alpha1/fun"
};
```

If you wanted people to access it at https://example.com/api/v1alpha1/fun you would specify proxy.api-prefix=/api/, and the gateway will do the rest.

Fixes #56

### DIFF
--- a/gwy/templates/main.go.tmpl
+++ b/gwy/templates/main.go.tmpl
@@ -44,11 +44,8 @@ func allowCors(cfg proxyConfig, handler http.Handler) http.Handler {
 
 // sanitizeApiPrefix forces prefix to be non-empty and end with a slash.
 func sanitizeApiPrefix(prefix string) string {
-  if len(prefix) == 0 {
-    prefix = "/"
-  }
-  if prefix[len(prefix)-1:] != "/" {
-    prefix = prefix + "/"
+  if len(prefix) == 0 || prefix[len(prefix)-1:] != "/" {
+    return prefix + "/"
   }
   return prefix
 }
--- a/gwy/templates/main.go.tmpl
+++ b/gwy/templates/main.go.tmpl
@@ -20,10 +20,18 @@ import (
 )
 
 type proxyConfig struct {
-  backend string
-  swagger string
+  // The backend gRPC service to listen to.
+  backend              string
+  // Path to the swagger file to serve.
+  swagger              string
+  // Value to set for Access-Control-Allow-Origin header.
   corsAllowOrigin      string
+  // Value to set for Access-Control-Allow-Credentials header.
   corsAllowCredentials string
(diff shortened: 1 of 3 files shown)


gwy/templates/main.go.tmpl
```diff
@@ -34,6 +42,17 @@ func allowCors(cfg proxyConfig, handler http.Handler) http.Handler {
   })
 }
 
+// sanitizeApiPrefix forces prefix to be non-empty and end with a slash.
+func sanitizeApiPrefix(prefix string) string {
+  if len(prefix) == 0 {
+    prefix = "/"
```
mhamrah: just `return "/"` here

gwy/templates/main.go.tmpl
```diff
@@ -34,6 +42,17 @@ func allowCors(cfg proxyConfig, handler http.Handler) http.Handler {
...
+  }
+  if prefix[len(prefix)-1:] != "/" {
+    prefix = prefix + "/"
```
mhamrah: `return fmt.Sprintf("%v/", prefix)`

gwy/templates/main.go.tmpl
```diff
@@ -49,7 +68,10 @@ func SetupMux(ctx context.Context, cfg proxyConfig) *http.ServeMux {
   if err != nil {
     log.Fatalf("Could not register gateway: %v", err)
   }
-  mux.Handle("/", allowCors(cfg, gwmux))
+
+  prefix := sanitizeApiPrefix(cfg.apiPrefix)
+  log.Println("API prefix is", prefix)
```
mhamrah: remove

gwy/templates/main.go.tmpl
mdkess: How's this?

gwy/templates/main.go.tmpl
mdkess: Fixed

PR #60, opened 2018-04-19
Remove duplicated SUPPORTED_LANGUAGES constant

blimmer:
This was defined and immediately overwritten with a newer constant.


bobbytables:
Thanks @blimmer ! Nice catch. Looks like a bad merge at some point.

PR #63, opened 2018-04-20
Remove max-depth option for directory flag.

### LINKED ISSUE #61: Directory flag specifies maxdepth 1
Today, pulling protofiles using the `-d` flag specifies a max-depth of 1.

https://github.com/namely/docker-protoc/blob/93b2381b888c691e2f4b3c0aac010f551cf361c2/all/entrypoint.sh#L174

Is there a specific reason for this? Can it be configurable?

We have a directory structure that looks like this:

```console
src/main/proto/company_pb
├── billing
│   └── events.proto
├── commons
│   ├── sqs
│   │   ├── sqs_attribute.proto
│   │   ├── sqs_attributes.proto
│   │   └── sqs_message.proto
│   └── timestamp.proto
```

So the `maxdepth` flag causes the command to fail out with this error
```
Missing input file.
```
because no files are passed to the underlying method.

blimmer:
Fixes #61.

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -170,8 +170,8 @@ PROTO_INCLUDE="-I /usr/include/ \
 
 if [ ! -z $PROTO_DIR ]; then
     PROTO_INCLUDE="$PROTO_INCLUDE -I $PROTO_DIR"
-    PROTO_FILES=(`find ${PROTO_DIR} -maxdepth 1 -name "*.proto"`)
-else 
+    PROTO_FILES=(`find ${PROTO_DIR} -name "*.proto"`)
+else
     PROTO_INCLUDE="-I . $PROTO_INCLUDE"
     PROTO_FILES=($FILE)
 fi


all/entrypoint.sh
```diff
@@ -170,8 +170,8 @@ PROTO_INCLUDE="-I /usr/include/ \
 
 if [ ! -z $PROTO_DIR ]; then
     PROTO_INCLUDE="$PROTO_INCLUDE -I $PROTO_DIR"
-    PROTO_FILES=(`find ${PROTO_DIR} -maxdepth 1 -name "*.proto"`)
-else 
+    PROTO_FILES=(`find ${PROTO_DIR} -name "*.proto"`)
```
blimmer: I tested this out locally and it worked just fine for my use-case where I have nested `.proto` files.

blimmer:
As discussed in the issue (#61), this could be considered a breaking change. So, a few thoughts:

Should this be released as 2.0?

If we didn't want to bump the major version, I could also make this customizable as a flag, defaulting to the `maxdepth 1` for now. If we went that route, I do think that changing the default to include all subdirectories is a more expected experience and should be released as a default in 2.x at some point.

Open to feedback on all of this 👍

mhamrah:
Thanks, will review and test within the next few days.

Re: versions. We've been trying this roughly to grpc versions, so when we bump to 1.11 I can add a note of this break change.

blimmer:
Sounds good - thanks @mhamrah . Let me know if more is needed from my end. Being able to use this container will be much easier than we do today! Thanks :-)

blimmer:
If anyone needs this in the meantime, feel free to use this image:

```
blimmer/protoc-all:pr-63
```


https://hub.docker.com/r/blimmer/protoc-all/tags/

mhamrah:
This looks good, going to merge. We're going to have to change some stuff on our side but it's for the best. I'm going to send a subsequent PR with a grpc upgrade and make a note this behavior changed.

blimmer:
Thank you! This project has helped us out a ton - thank you for open-sourcing it and being so responsive to issues and PRs 🥇

mhamrah:
No problem, thanks for the contrib!

PR #65, opened 2018-04-25
Properly support CORS and preflight checks.

mdkess:


gwy/templates/config.yaml.tmpl
```diff
@@ -2,6 +2,7 @@ backend:
 cors:
   allow-origin:
   allow-credentials:
+  allow-methods:
```
mhamrah: Should we set this as a default?

PR #68, opened 2018-04-28
Support java outputting to a jar.

### LINKED ISSUE #66: Passing a `.jar` file as the output produces a directory
Conveniently, the java_out can handle a `.jar` extension and automatically jar up the `.java` files for you.

https://developers.google.com/protocol-buffers/docs/reference/java-generated#invocation

> When outputting Java code, the protocol buffer compiler's ability to output directly to JAR archives is particularly convenient, as many Java tools are able to read source code directly from JAR files. To output to a JAR file, simply provide an output location ending in .jar. Note that only the Java source code is placed in the archive; you must still compile it separately to produce Java class files.

However, the entrypoint always assumes that the `-o` parameter is a directory:

https://github.com/namely/docker-protoc/blob/e9304e8b05d68256584bf848c8a1cd48af7ccefc/all/entrypoint.sh#L125-L127

So if you invoke with a `jar` parameter:

```console
-o build/java/my.jar
```

It will fail with this error message:

```
Generating java files for proto in build/java/foo.jar
build/java/my.jar: Is a directory%
```

I think the solution is to check and only `mkdir -p` through to the nearest directory here:

https://github.com/namely/docker-protoc/blob/e9304e8b05d68256584bf848c8a1cd48af7ccefc/all/entrypoint.sh#L125-L127

mdkess:
Fixes #66

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -123,7 +123,13 @@ fi
 echo "Generating $GEN_LANG files for ${FILE}${PROTO_DIR} in $OUT_DIR"
 
 if [[ ! -d $OUT_DIR ]]; then
-  mkdir -p $OUT_DIR
+  # If a .jar is specified, protoc can output to the jar directly. So
+  # don't create it as a directory.
+  if [[ "$GEN_LANG" == "java" ]] && [[ $OUT_DIR == *.jar ]]; then
+    mkdir -p `dirname $OUT_DIR`
+  else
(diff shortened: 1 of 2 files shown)


blimmer:
Thank you!

mdkess:
@blimmer Thanks for your contributions and feature requests! We're happy to help - just open issues and we'll add them 👍

PR #69, opened 2018-05-09
Expose linting

blimmer:
My team is working on introducing linting of our protobufs. I noticed that `protoc-gen-lint` is being installed in the container:

https://github.com/namely/docker-protoc/blob/32cda135bca79697425b25f67f4aa69ceadfc85a/all/Dockerfile#L39

But it's not being exposed via the entrypoint. This PR adds a flag to enable linting. I've tested this on my project and it's working as I'd expect.

This is available under my dockerhub if you want to test it out:

```
blimmer/protoc-all:1.11-lint
```

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -14,11 +14,14 @@ printUsage() {
     echo " -l LANGUAGE          The language to generate (${SUPPORTED_LANGUAGES[@]})"
     echo " -o DIRECTORY         The output directory for generated files. Will be automatically created."
     echo " -i includes          Extra includes"
+    echo " --lint CHECKS        Enable linting protoc-lint (CHECKS are optional - see https://github.com/ckaznocha/protoc-gen-lint#optional-checks)"
     echo " --with-gateway       Generate grpc-gateway files (experimental)."
 
 }
 
 GEN_GATEWAY=false
+LINT=false
+LINT_CHECKS=""
 SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc" "gogo" "php" "node")
 EXTRA_INCLUDES=""
 OUT_DIR=""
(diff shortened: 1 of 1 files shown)


all/entrypoint.sh
```diff
@@ -159,6 +169,15 @@ plugins=grpc+embedded\
         ;;
 esac
 
+LINT_STRING=''
+if [[ $LINT == true ]]; then
+    if [[ $LINT_CHECKS == '' ]]; then
+        LINT_STRING="--lint_out=."
+    else
+        LINT_STRING="--lint_out=$LINT_CHECKS:."
+    fi
```
blimmer: There might be a better way to handle this inner if statement, but I'm no bash expert 😄

all/entrypoint.sh
```diff
@@ -71,6 +74,13 @@ while test $# -gt 0; do
             GEN_GATEWAY=true
             shift
             ;;
+        --lint)
+            LINT=true
+            if test $# -gt 1; then
```
mdkess: This is going to take the next arg no matter what - i.e. if we say `--lint --with-gateway`, it'll set `LINT_CHECKS=--with-gateway`. Also you'll need another `shift` inside the `if` block, or you'll re-process the next command.

I think what you'll want to do is say `if [[ $2 == -* ]]; then`, i.e. check that the next arg starts with a `-`.

blimmer: Ah, yes you're right. Thank you for catching that. Working on a fix now.

all/entrypoint.sh
```diff
@@ -71,6 +74,13 @@ while test $# -gt 0; do
             GEN_GATEWAY=true
             shift
             ;;
+        --lint)
+            LINT=true
+            if [ "$#" -gt 1 ] && [[ $2 != -* ]]; then
+                LINT_CHECKS=$2
```
mdkess: `shift` here too (i.e. consume the $2), otherwise everything looks good.

I agree - the bash is getting kinda overboard, it's on my short-list to port this to a more sane language.

blimmer:
@mdkess this is ready to go :-)

mdkess:
Awesome, thanks! I'll build and push now.

mdkess:
Should be pushed as namely/protoc-all:1.11 - thanks again for the contribution 💯

PR #73, opened 2018-05-16
Allow concatenating includes

### LINKED ISSUE #72: Allow passing multiple "includes"
As implemented today, you can't pass multiple includes via a `-i` parameter. Each invocation of `-i` overwrites the last.

Here's the code that does that:

https://github.com/namely/docker-protoc/blob/3a872063daabf159aaed36a38d8467e2d00a1913/all/entrypoint.sh#L69-L72

So, invocations like this don't work

```
docker run ... namely/protoc-all:1.11 -i protorepo -i something_else
```

It would be nice if each `-i` added onto the `EXTRA_INCLUDES`.

blimmer:
Fixes #72

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -67,7 +67,7 @@ while test $# -gt 0; do
             shift
             ;;
         -i) shift
-            EXTRA_INCLUDES="-I$1"
+            EXTRA_INCLUDES="$EXTRA_INCLUDES -I$1"
             shift
             ;;
         --with-gateway)


all/entrypoint.sh
```diff
@@ -67,7 +67,7 @@ while test $# -gt 0; do
             shift
             ;;
         -i) shift
-            EXTRA_INCLUDES="-I$1"
+            EXTRA_INCLUDES="$EXTRA_INCLUDES -I$1"
```
blimmer: this will add an extra space for the first include, but it's the cleanest logic without an additional `if` statement IMO. open to a more complex implementation if anyone feels strongly about this

blimmer:
I pushed this change to dockerhub if you'd like to test it out. the tag is:

```
blimmer/protoc-all:1.11-extra-includes
```

blimmer:
cc @mdkess since we worked on a similar problem with #69

mdkess:
Nice, thanks!

PR #76, opened 2018-06-14
Make gateway binary totally static

johnnymo87:
I experimenting with running this binary from the same container as my gRPC server. If that turns out to be a good idea, I need this binary to be totally static so that it works anywhere.


mhamrah:
Closing this. We can revisit.

PR #86, opened 2018-08-08
Upgrade grpc to 1.14

bobbytables:
Also include swagger protos in include so you can easily use them in proto files.

@kaimou1357 @mdkess 

It's super weird the way we have to use swagger annotations. If they're in the proto they should be available just like the Google proto types. This includes the swagger annotations proto in the include path for proto so you can use them easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ If you're having trouble, see [Docker troubleshooting](#docker-troubleshooting) 
 Pull the container:
 
 ```sh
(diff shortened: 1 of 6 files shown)


mhamrah:
The gwy may also benefit from a better multi-stage build in the /all directory where we target different layers via make

bobbytables:
Done. We need to create a real build system for this. CircleCI has good OpenSource support for docker builds now in their v2.

PR #87, opened 2018-08-20
Refactor all container to prep for consolidation, removes duplication

mhamrah:
This is the first phase of a restructure to consolidate all dockerfiles to a single-multi stage build and remove duplication. It relies on a new root-level build script, `build.sh`, which uses docker args to pass grpc version information to lower containers. It also introduces a new naming convention to separate out gRPC versions from container versions (see readme).


bobbytables:
Nice, this is cool.

PR #89, opened 2018-08-22
move gateway into unified docker file

mhamrah:
This is the last step to unify the various dockerfiles, move the gen-grpc-gateway container build into the root-level container.


bobbytables:
Neat.

PR #96, opened 2018-10-17
Splitting up grpc and grpc-java

shiena:
Since grpc is cloned from the branch, it is not clear what commit it was built from.
Therefore, it seems better to build from the release tag instead of the branch.
Also, since grpc and grpc-java release versions are sometimes different, they are distinguished.


mhamrah:
I like the idea of splitting up grpc and java, but don't think we should track the docker container to the patch version.

shiena:
OK. I changed it only to split grpc and java.

PR #99, opened 2018-11-14
Put header logging behind a config flag

aldelucca1:
This will make it so request headers are no longer logged by default. To enable header logging you can set:
```
log_headers = true
```

### DIFF
--- a/gwy/templates/main.go.tmpl
+++ b/gwy/templates/main.go.tmpl
@@ -29,6 +29,8 @@ type proxyConfig struct {
 	backend string
 	// The log level to use
 	logLevel string
+	// Whether to log request headers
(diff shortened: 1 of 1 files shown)


aldelucca1:
@mhamrah how do you see that working, would you see an additional log statement put out at DEBUG with just the headers? Right now the primary log line is put out at INFO.  So the 2 options I saw was to either do it with an additional flag to enrich the current INFO level log, or put the headers out into a separate log statement at DEBUG.

mhamrah:
Ah, i see. I was thinking some function like `logRequest(req)` that if level == debug, you include the headers in a level.Info, but if it was info, you don't include the headers. Might be too much work for this.

PR #104, opened 2019-01-13
Added cpp to list of languages

MyCADDev:
Added cpp to list of languages

I did not add any tests but I tested manually and it produces the expected cpp code.


MyCADDev:
Just added and pushed to same pull request branch

Harish

mhamrah:
thanks! will build and push now

PR #108, opened 2019-02-12
make scripts fail on error

mdkess:


all/test/test.proto
```diff
@@ -1,9 +1,9 @@
 syntax = "proto3";
 
-package Messages;
-
 import "google/api/annotations.proto";
-import "include.proto";
+import "all/test/include.proto";
```
mdkess: I had to change this to get it to generate valid Go - I wonder if we haven't run the tests in a while.

PR #116, opened 2019-04-03
Adding option to configure response headers from env vars

ido-namely:
The proposed changes would enable configuring additional headers as part of each http response provided by the grpc-gateway.

It would help with [this](https://namely.atlassian.net/browse/PU-3806) bug by sending 
`Cache-Control: no-cache` in the header.
The env var set in the config map of the service would be in this case:
`ORGUNITSREADER_RESPONSE-HEADERS_CACHE-CONTROL="no-cache"`

### DIFF
--- a/gwy/templates/main.go.tmpl
+++ b/gwy/templates/main.go.tmpl
@@ -26,6 +26,8 @@ import (
 )
 const (
 	ResponseHeaders = "RESPONSE-HEADERS"
+	EnvVarSepChar = "_"
+	ViperKeySepcChar = "."
 )
 type proxyConfig struct {
 	// The backend gRPC service to listen to.
@@ -266,9 +268,9 @@ func SetupMux(ctx context.Context, cfg proxyConfig) *http.ServeMux {
 func bindEnvVars(viper *viper.Viper, r *strings.Replacer, prefix string, secondPrefix string) {
 	for _, e := range os.Environ() {
 		pair := strings.Split(e, "=")
-		if strings.HasPrefix(pair[0], prefix+"_") {
-			key := strings.TrimPrefix(pair[0], prefix+"_")
-			if strings.HasPrefix(key, secondPrefix+"_") {
+		if strings.HasPrefix(pair[0], prefix+EnvVarSepChar) {
+			key := strings.TrimPrefix(pair[0], prefix+EnvVarSepChar)
+			if strings.HasPrefix(key, secondPrefix+EnvVarSepChar) {
 				viper.BindEnv(strings.ToLower(r.Replace(key)), pair[1])
 			}
 		}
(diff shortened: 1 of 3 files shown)


gwy/templates/main.go.tmpl
```diff
@@ -206,6 +215,21 @@ func outgoingHeaderMatcher(metadata string) (string, bool) {
 	return metadata, true
 }
 
+// getResponseHeaders will identified keys and values which are response headers settings and 
+// translate into a map
+func getResponseHeaders(cfg *viper.Viper) map[string]string {
+	m := make(map[string]string)
+	headersPrefix := strings.ToLower(ResponseHeaders)
+	for _, key := range cfg.AllKeys() {
+		if strings.HasPrefix(key, headersPrefix) {
```
mdkess: I think should be `strings.HasPrefix(key, headersPrefix + ".")`

gwy/templates/main.go.tmpl
```diff
@@ -234,10 +258,23 @@ func SetupMux(ctx context.Context, cfg proxyConfig) *http.ServeMux {
 
 	prefix := sanitizeApiPrefix(cfg.apiPrefix)
 	logrus.Infof("API prefix is: %s", prefix)
-	mux.Handle(prefix, handlers.CustomLoggingHandler(os.Stdout, http.StripPrefix(prefix[:len(prefix)-1], allowCors(cfg, gwmux)), formatter))
+	mux.Handle(prefix, handlers.CustomLoggingHandler(os.Stdout, http.StripPrefix(prefix[:len(prefix)-1], addRespHeaders(cfg, gwmux)), formatter))
 	return mux
 }
 
+// bindEnvVars will bind all env vars that starts with "prefix_secondPrefix"
+func bindEnvVars(viper *viper.Viper, r *strings.Replacer, prefix string, secondPrefix string) {
```
mdkess: I'd combine prefix and secondPrefix in the API - i.e. do the concat as a caller, it'll be easier.

ido-namely: good call

ido-namely: actually I need to to remove use both prefix for identifying the keys but than trim only the first prefix, so I think it be better to leave it as is.

gwy/templates/main.go.tmpl
```diff
@@ -206,6 +217,21 @@ func outgoingHeaderMatcher(metadata string) (string, bool) {
 	return metadata, true
 }
 
+// getResponseHeaders will identified keys and values which are response headers settings and 
+// translate into a map
+func getResponseHeaders(cfg *viper.Viper) map[string]string {
+	m := make(map[string]string)
+	headersPrefix := strings.ToLower(ResponseHeaders)
+	for _, key := range cfg.AllKeys() {
+		if strings.HasPrefix(key, headersPrefix+".") {
```
aldelucca1:
```suggestion
		if strings.HasPrefix(key, headersPrefix+ViperKeySepcChar) {
```

gwy/templates/main.go.tmpl
```diff
@@ -206,6 +217,21 @@ func outgoingHeaderMatcher(metadata string) (string, bool) {
...
+			value := cfg.GetString(key)
+			header := strings.TrimPrefix(key, headersPrefix+".")
```
aldelucca1:
```suggestion
			header := strings.TrimPrefix(key, headersPrefix+ViperKeySepcChar)
```

PR #120, opened 2019-04-29
flag to not include google protos

mdkess:
this is useful if one wants to use local versions of these protos.


mhamrah:
this is probably because I generate the google protos in /usr/local/include during the build step. Instead of moving the whole include directory, we can generate those files in something like `/opt/` and modify this default script to include `/opt/` by default and do nothing when `no-google-includes` is true

Issue #111, opened 2019-02-21
Feature Request: Add typescript definition support for node (node-ts) with protoc-gen-ts plugin


HectorFriedman-Cintron:
It would be nice to add typescript support for node via the protoc-gen-ts plugin.

I'm happy to take this on as a PR.

Potential Changes to Dockerfile
```
RUN set -ex && apk --update --no-cache add \
 ...
    nodejs \
    nodejs-npm \
    && npm i -g ts-protoc-gen

ENV PROTOC_GEN_TS_PATH /usr/local/lib/node_modules/protoc-gen-ts
```

Potential Changes to all/Entrypoint.sh
```
case $GEN_LANG in
...
"node(-ts)*") ...
...
esac
....
if [[ $GEN_LANG == "node-ts" ]]; then
    protoc \
        --plugin="protoc-gen-ts=${PROTOC_GEN_TS_PATH}" \
        --ts_out=$OUT_DIR \
        ${PROTO_FILES[@]}
fi
```

damiannolan:
This would be awesome!

esilkensen:
I'd like to see this too 👍

Are you referring to [`ts-protoc-gen`](https://github.com/improbable-eng/ts-protoc-gen) (rather than `protoc-gen-ts`)?

mhamrah:
PR's welcome!

esilkensen:
Sounds good, I will give this a shot and hope to open a PR!

mhamrah:
Believe we can close this now that's to #125 !

PR #125, opened 2019-05-15
Add TypeScript declaration support using protoc-gen-ts plugin

esilkensen:
This PR is an implementation of the feature request described in https://github.com/namely/docker-protoc/issues/111:

My proposal is to add a `--with-typescript` option to the `node` language that leverages the [ts-protoc-gen](https://github.com/improbable-eng/ts-protoc-gen#readme) plugin to generate TypeScript declaration files (`.d.ts` files) that match the JavaScript output of `protoc --js_out=import_style=commonjs,binary`.

That plugin can also generate service definitions that include TypeScript declarations in the structure required by the [Improbable grpc-web implementation](https://github.com/improbable-eng/grpc-web). This PR doesn't use that feature; I thought if it would be useful to expose it as an option, maybe it would make sense as an option to the `web` language, and could be its own PR?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,7 +60,13 @@ RUN set -ex && apk --update --no-cache add \
     bash \
     libstdc++ \
     libc6-compat \
-    ca-certificates
+    ca-certificates \
+    nodejs \
+    nodejs-npm
+
+# Add TypeScript support
+
+RUN npm i -g ts-protoc-gen@0.10.0
 
 COPY --from=build /tmp/grpc/bins/opt/grpc_* /usr/local/bin/
 COPY --from=build /tmp/grpc/bins/opt/protobuf/protoc /usr/local/bin/
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -15,6 +15,7 @@ printUsage() {
     echo " --lint CHECKS        Enable linting protoc-lint (CHECKS are optional - see https://github.com/ckaznocha/protoc-gen-lint#optional-checks)"
     echo " --with-gateway       Generate grpc-gateway files (experimental)."
     echo " --with-docs FORMAT   Generate documentation (FORMAT is optional - see https://github.com/pseudomuto/protoc-gen-doc#invoking-the-plugin)"
+    echo " --with-typescript    Generate TypeScript declaration files (.d.ts files) - see https://github.com/improbable-eng/ts-protoc-gen#readme"
     echo " --go-source-relative Make go import paths 'source_relative' - see https://github.com/golang/protobuf#parameters"
     echo " --no-google-includes Don't include Google protobufs"
 }
@@ -23,6 +24,7 @@ printUsage() {
 GEN_GATEWAY=false
 GEN_DOCS=false
 DOCS_FORMAT="html,index.html"
+GEN_TYPESCRIPT=false
 LINT=false
 LINT_CHECKS=""
 SUPPORTED_LANGUAGES=("go" "ruby" "csharp" "java" "python" "objc" "gogo" "php" "node" "web" "cpp" "descriptor_set")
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -60,7 +60,13 @@ RUN set -ex && apk --update --no-cache add \
...
     libstdc++ \
     libc6-compat \
-    ca-certificates
+    ca-certificates \
+    nodejs \
+    nodejs-npm
+
+# Add TypeScript support
+
+RUN npm i -g ts-protoc-gen google-protobuf
```
esilkensen: The latest release of `ts-protoc-gen` is missing `google-protobuf` as a runtime dependency. This was fixed in April (https://github.com/improbable-eng/ts-protoc-gen/pull/161) on master but there hasn't been a new release since January (0.9.0).

all/entrypoint.sh
```diff
@@ -184,6 +186,9 @@ plugins=grpc+embedded\
     "node")
         GEN_STRING="--grpc_out=$OUT_DIR --js_out=import_style=commonjs,binary:$OUT_DIR --plugin=protoc-gen-grpc=`which grpc_${PLUGIN_LANG}_plugin`"
         ;;
+    "node-ts")
+        GEN_STRING="--grpc_out=$OUT_DIR --js_out=import_style=commonjs,binary:$OUT_DIR --ts_out=$OUT_DIR --plugin=protoc-gen-grpc=`which grpc_${PLUGIN_LANG}_plugin`"
```
esilkensen: the ts-protoc-gen plugin also supports generating gRPC stubs using grpc-web: https://github.com/improbable-eng/ts-protoc-gen#generating-grpc-service-stubs-for-use-with-grpc-web

I wondered if `node-ts` should drop the `protoc-gen-grpc` plugin and instead include the `service=true` parameter in `ts_out`?

jonny-improbable: Note that there are 2 'grpc-web' implementations: Improbable's and Googles. ts-protoc-gen only generates stubs for use with Improbable's implementation.

esilkensen: Thanks Jonny, and thanks for the new release!

So if I understand correctly, what’s in this PR would generate TypeScript declarations for protobuf messages and Google grpc-web service stubs without TypeScript declarations.

Then a second option could be to instead generate Improbable grpc-web service stubs with TypeScript declarations.

I’m curious if you have a recommendation?

Seems like it could be nice to fit in with the rest of this project using Google grpc-web, but that it’d also be nice for “node-ts” to generate TypeScript declarations for all the generated code?

(I was just reading https://github.com/improbable-eng/grpc-web/issues/199 and https://grpc.io/blog/state-of-grpc-web)

Dockerfile
```diff
@@ -60,7 +60,13 @@ RUN set -ex && apk --update --no-cache add \
...
     libstdc++ \
     libc6-compat \
-    ca-certificates
+    ca-certificates \
+    nodejs \
+    nodejs-npm
+
+# Add TypeScript support
+
+RUN npm i -g ts-protoc-gen@0.10.0
```
esilkensen: I wondered if it's OK to install `nodejs` and `ts-protoc-gen` here, or if they should be moved to the `build` stage and copied following the examples below?

mhamrah: i'm fine with it here; i think it needs to be here to make running the `node` command easier in the runtime container.

all/entrypoint.sh
mhamrah: I'd be open to add explicit options for node. That way, you'd have `-l node` and something like `--with-typescrypt` `--with-improbable` or `--with-ts-protoc-gen`

Thoughts?

esilkensen: Nice, yeah I like that 👍

esilkensen: I tried adding a `--with-typescript` option that only applies to `node` (and updated the PR description). Thanks for that suggestion - let me know what you think, whenever you have a chance to check it out!

mhamrah:
Thanks sorry for the delay!

PR #126, opened 2019-05-29
Add descriptor_set_out flags to command line tool

### LINKED ISSUE #124: descriptor_set_out does not allow for `--include_imports` and `--include_source_info`
The descriptor set out is a good option but to take full advantage of the functionality it would most helpful to be able to set the --include_imports flag and the --include_source_info flag.

tsexton6689:
Fixes #124


mhamrah:
Thanks!

PR #137, opened 2019-07-30
Update google paths in readme

charandas:
I looked up the actual paths in current dockerfile, and tested with a proto which uses WKTs. The paths are no longer under `/usr`. They are under `/opt`.


mhamrah:
Thanks for this! @mdkess do you remember moving these though? Wondering if we should symlink/move back?

PR #147, opened 2019-11-09
Update grpc-web to 1.0.7

Sovietaced:
My team would like to pick up some changes from the latest `grpc-web` release.


Sovietaced:
Thanks for the quick turnaround. Any chance this can warrant a minor/patch release or is there a release cadence I can wait for?

PR #150, opened 2019-11-13
Add support to go-proto-validators

cristianovagos:
Added support to `go-proto-validators` from https://github.com/mwitkow/go-proto-validators


cristianovagos:
@mhamrah forgot to add `COPY --from=build /go/src/github.com/mwitkow/go-proto-validators/ /opt/include/github.com/mwitkow/go-proto-validators/` into Dockerfile. 
I will push it asap

PR #154, opened 2019-11-14
Adds support to pass the csharp options to protoc

abe545:
This is needed, because there are some [options for csharp codegen](https://developers.google.com/protocol-buffers/docs/reference/csharp-generated#compiler_options) that we're going to need. Specifically, we need to set `base_namespace`, but I feel like this is better than trying to break the individual options out.


all/entrypoint.sh
```diff
@@ -256,6 +263,11 @@ plugins=grpc+embedded\
             GEN_STRING="$GEN_STRING --include_source_info"
         fi
         ;;
+    "chsarp")
```
k-ruddy: typo?

abe545: Yeah - already pushed a fix :)

PR #159, opened 2019-12-24
Set grpc version 1.26

nikita-namely:


variables.sh
```diff
@@ -2,7 +2,7 @@
 
 DOCKER_REPO=${DOCKER_REPO}
 NAMESPACE=${NAMESPACE:-namely}
-GRPC_VERSION=${GRPC_VERSION:-1.25}
+GRPC_VERSION=${GRPC_VERSION:-1.26}
 GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.24}
```
ido-namely: wondering if we should update this now as well?

variables.sh
```diff
@@ -2,7 +2,7 @@
...
 BUILD_VERSION=${BUILD_VERSION:-0}
```
ido-namely: I had just merged another PR, please make sure to rebase and change this value back to 0 (I had to change it to 1)

PR #165, opened 2020-01-30
Add support for Scala with ScalaPBC

spoonscen:
This PR adds support for scala using the ScalaPBC as a protoc plugin.
https://scalapb.github.io/scalapbc.html

It also rolls back the version bump that is currently breaking the master build.


Dockerfile
```diff
@@ -56,8 +57,14 @@ RUN make -C /go/src/github.com/envoyproxy/protoc-gen-validate/ build
 
 RUN go get -u github.com/mwitkow/go-proto-validators/protoc-gen-govalidators
 
-# Add grpc-web support
 
+
+# Add scala support
```
spoonscen: Do we want to make this one `RUN` command? I saw both multiple `RUN` commands as well as `&&` commands in the `Dockerfile`.

aldelucca1:
Thanks for this, sorry for the broken master, all is good now, if you wouldn’t mind rebasing your change onto the current master it’d be greatly appreciated

Dockerfile
aldelucca1: Yeah if we could collapse this to a single RUN it will generate less layers

spoonscen:
@aldelucca1 Thanks for the review! I rebased and collapsed into a single `RUN`.

spoonscen:
@aldelucca1 Anything else you need from me to get this moved along? Thanks!

spoonscen:
🎉

PR #166, opened 2020-02-07
Added support for nanopb

joaquimjmsilva:
This PR adds support for nanopb protoc plugin.
https://jpa.kapsi.fi/nanopb/

Nanopb is a plain-C implementation that targets microcontrollers with tight memory constraints.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ FROM golang:$go-alpine$alpine AS build
 ARG grpc
 ARG grpc_java
 ARG grpc_web=1.0.7
+ARG nanopb=0.4.1
 
 RUN set -ex && apk --update --no-cache add \
     bash \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -70,7 +76,10 @@ RUN set -ex && apk --update --no-cache add \
     libc6-compat \
     ca-certificates \
     nodejs \
-    nodejs-npm
+    nodejs-npm \
+    python \
+    py2-pip && \
+    pip install protobuf
```
ido-namely, line 82: can you explain why the python packages are necessary?
does the plugin require them? because looking at the download page it seems like all the dependencies are included, no?

Additionally, I wonder if we should specify the version of this packages.

ido-namely (commented):
@joaquimjmsilva thank you for this contribution, apologies for the major delay in our response.

If you're still interested in seeing this PR merged, will you mind adding tests?

Thanks again, much appreciated!

PR #175, opened 2020-04-08
Add swagger json name option

briantenazas:
- Adds a flag to protoc-gen-swagger to have the generated swagger file use JSON names (camelCased) instead of the raw proto field names, which are usually snake_cased.


briantenazas:
@aldelucca1 can you take a look when you get a chance please?

aldelucca1:
@briantenazas doh sorry this took so long, if you could just bump the build number in `vars.sh` we can get this merged

briantenazas:
@aldelucca1  done!

all/entrypoint.sh
```diff
@@ -172,6 +174,10 @@ while test $# -gt 0; do
             CSHARP_OPT=$1
             shift
             ;;
+        ----with-swagger-json-names)
```
aldelucca1, line 177: Think there are actually too many `-` here, should probably be
```suggestion
        --with-swagger-json-names)
```

briantenazas: thanks!

briantenazas: @aldelucca1  should be all set. passes build check

PR #178, opened 2020-05-08
Add support for kroto-plus (Kotlin/JVM extensions)

iainmcgin:
The kroto-plus plugin produces some extensions to the Java generated types that simplify Kotlin code interactions with protobuf and gRPC.


iainmcgin:
Closing due to lack of response

PR #185, opened 2020-07-31
Updated base golang version of gateway, and updated to use standardized build args

abe545:
The current image that we use is 2 years old, and we can't specify the golang version, so we can't even pass args to build it with newer versions.
I updated the args to match the ones we use in the other images.


gwy/templates/Dockerfile.tmpl
```diff
@@ -1,7 +1,8 @@
 # Template for grpc-gateway 
 
-ARG alpine_version=3.8
-FROM golang:1.10-alpine$alpine_version AS build
+ARG alpine=3.11
+ARG go=1.12
```
vtopc, line 4: The current stable version is `1.14`. Why `1.12`?

abe545: I'm matching what is already used in the other images in this repo. My goal is to get everything on the same version. We can update them all together in the (hopefully near) future.

vtopc: OK. Sorry.

gwy/templates/Dockerfile.tmpl
```diff
@@ -18,7 +19,7 @@ WORKDIR /app
 # Build the gateway
 RUN go build -o grpc_gateway src/pkg/main/main.go
 
-FROM alpine:$alpine_version
+FROM alpine:$alpine
```
vtopc, line 22: Could you keep old naming(`alpine_version`)?

abe545: I could... but all the other containers use `alpine`. I think this will be better long term if everything matches.

abe545: I guess this is the only one that is a generated dockerfile for tho. Okay, I'll change it back

vtopc: I would say, that better to change other containers to be more explicit.

PR #187, opened 2020-08-05
Fix path to gateway's Dockerfile 

vtopc:
In the #186, I broke backward compatibility with path to gateway's Dockerfile.
This is the fix.


vtopc:
fixed in the protorepo

PR #188, opened 2020-08-13
Update grpc-web to v1.2.0

pzeinlinger:
grpc-web 1.2.0 includes a lot of refinements, improvements, and bug fixes. The build has been tested and make test returns no error.


abe545:
Thanks @pzeinlinger! I want to push grpc 1.31 out too - will include this change with that.

abe545:
Merged as part of #189

PR #189, opened 2020-08-19
Release grpc 1.30

abe545:
I think we should also release 1.31, but want to make sure we have this first.


vtopc:
Does this mean, that we need to switch to `google.golang.org/grpc@v1.30` in Golang too?

abe545:
> Does this mean, that we need to switch to `google.golang.org/grpc@v1.30` version in Golang too?

I don't see that anywhere in this repo - we just use `google.golang.org/grpc` without a version.

vtopc:
> > Does this mean, that we need to switch to `google.golang.org/grpc@v1.30` version in Golang too?
> 
> I don't see that anywhere in this repo - we just use `google.golang.org/grpc` without a version.

No. I'm asking about go-commo(https://github.com/namely/go-common/blob/master/go.mod#L50) and other services.

abe545:
> > > Does this mean, that we need to switch to `google.golang.org/grpc@v1.30` version in Golang too?
> > 
> > 
> > I don't see that anywhere in this repo - we just use `google.golang.org/grpc` without a version.
> 
> No. I'm asking about go-commo(https://github.com/namely/go-common/blob/master/go.mod#L50) and other services.

We will when we update to this version in the protorepo (and your service updates the protorepo reference)

vtopc:
OK. Looks good in go-common test - https://github.com/namely/go-common/pull/343

vtopc (approved):

variables.sh
```diff
@@ -2,12 +2,12 @@
...
-GRPC_VERSION=${GRPC_VERSION:-1.29}
-GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.29}
-GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.0.7}
+GRPC_VERSION=${GRPC_VERSION:-1.30}
+GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.30}
+GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.2.0}
 GO_VERSION=${GO_VERSION:-1.14}
 ALPINE_VERSION=${ALPINE_VERSION:-3.12}
-BUILD_VERSION=${BUILD_VERSION:-4}
+BUILD_VERSION=${BUILD_VERSION:-1}
```
ido-namely, line 10: I think this can be reset to 0

PR #190, opened 2020-08-19
Release grpc 1.31

abe545:
So, the update to the build system in grpc 1.31 broke our builds. I tried my hardest to figure it out, but cmake isn't my forte. 

The good news is that most of it is hosted in APK for alpine - I updated to pull those packages. One thing to note about this approach: the `edge` channel which has the latest & greatest versions of grpc won't keep the packages around forever. So if we want to do it this way, we're going to have to stay on top of grpc/protobuf releases, and push them out ourselves.

If you don't like using the packages, here is the [APK build](https://git.alpinelinux.org/aports/tree/community/grpc/APKBUILD) to get you started.


vtopc (approved):

abe545:
@ido-namely @vtopc this isn't done - the binaries need to be sorted in the final image, but before I finished the work, I wanted to check with you about the change in direction for how we get the binaries in the image.

ido-namely:
> @ido-namely @vtopc this isn't done - the binaries need to be sorted in the final image, but before I finished the work, I wanted to check with you about the change in direction for how we get the binaries in the image.

@abe545  I think this is the right move. it would significantly help with maintaining this repo. I appreciate you working through this.

vtopc:
Sorry, cmake isn't my forte too :(

mhamrah:
https://github.com/grpckit/grpckit/blob/master/Dockerfile#L31

abe545:
> https://github.com/grpckit/grpckit/blob/master/Dockerfile#L31

Thanks @mhamrah! I see you guys switched to the debian base image - would probably make all this easier... The grpc documentation suggests using bazel to build, but that doesn't have packages available for alpine, so the debian image would make using their favored build system easier 🤷

vtopc (approved):

mhamrah:
Yeah I was fighting for a bit, and gave up. I like the Debian slim images.

PR #191, opened 2020-08-26
Try to build grpc ourselves still and upgrade to 1.31

abe545:
In #190, I got the binaries from APK that I could. This is attempting to build it from source. Since the grpc cli is now only built when we build the tests, I'm not sure this is going to complete within the pipeline timeout of 60s. I'm thinking the APK approach is easier to maintain for us, but wanted to give this option if we think it is better.


abe545:
Yeah, we hit the 60 minute cap for free builds in azure pipelines with this approach (grpc got about 80% built according to logs). Since this is a public project, I think we can ask microsoft to increase the limit, but the question in my mind is if we want to spend any more time on this. My vote is no... @ido-namely @vtopc what do you think?

vtopc:
@abe545, let's choose solution, which is easier to maintain(APK?). 
And about 60 minutes cap - we could switch to the GitHub Actions(looks like there is 6 hours limit - https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits)

abe545:
> @abe545, let's choose solution, which is easier to maintain(APK?).
> And about 60 minutes cap - we could switch to the GitHub Actions(looks like there is 6 hours limit - https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits)

This is a github action! But the APK solution builds in 6 minutes, so I don't think we need to do anything else to upgrade to 1.31

PR #196, opened 2020-09-03
change validate_out=lang value to go

jcorry:
I think the value for `--validate_out=lang=` should be `go`, not `gogo`, right?

When I use this in `namely/protoc-all` I get the error: 
```
[validator]could not find templates for `lang`: gogo
--validate_out: protoc-gen-validate: Plugin failed with status code 1.
```

I am looking at the code in `validate.go` and it looks for a template matching the value passed in this param...and there is no template for `gogo`.

I think this change should fix this.


jcorry:
I'm not sure why the pipeline is failing...nothing I changed should have affected that, right?

gebv:
```
[validator]could not find templates for `lang`: gogo
--validate_out: protoc-gen-validate: Plugin failed with status code 1.
make: *** [gen] Error 1
```
@jcorry it is problems show up after version `1.29_1`
This is repo without tags - couldn’t find what broke in the code. Did you find the problem?

jcorry:
Hi, @gebv I don't see anything that should cause the build to fail. I was hoping to contribute a helpful change/fix :(

gebv:
I wanted to say - this problem has long been (after `1.29_1`). 
@abe545 could you clarify what has changed after version `1.29_1`?

gebv:
My code
```bash
PROTOC_MAP ?= Mch-common/types.proto=github.com/my/shared/proto/files,Mch-common/shared_storage.proto=github.com/my/shared/proto/files

PROTOC ?= docker run --rm -u ${shell id -u} \
	-v ${PWD}:/defs namely/protoc-all:1.29_1 \
	-l gogo \
	--go-package-map ${PROTOC_MAP} \
	--with-validator \
	-o ./ \
	-i ./proto

$(PROTOC) \
		-d ./proto/path/to/proto/file
```

abe545:
@jcorry @gebv -
We switched the build to use the alpine package manager to get the `grpc`, `grpc-cli`, and `protoc` binaries - this is because of a change to the grpc repo, where now the `grpc-cli` will only build if you build the test suite too. And that was taking > 60 minutes to build, timing out our build server. Unfortunately with the way that alpine manages the packages, they drop versions of packages from the edge channel (where the actual versions of these binaries live that we need) as new ones are pushed.

I think the only way for us to have repeatable builds at this point is to switch to a distro that has stable packages.

ido-namely:
Hi @jcorry, @gebv, thank you for the contribution and interest.
IS there any chance you'd still be interested in contributing further by adding a test to this PR?
Please note that we're currently on gRPC version 1.42.
Thanks again!

ido-namely:
superseded by https://github.com/namely/docker-protoc/pull/326

PR #197, opened 2020-09-03
Passthrouge protoc options via --protoc-opts

vincenthcui:
Usage:
	`$CMD --protoc-opts --go_out=:.`


ido-namely:
I don't think this is necessary given the several flag options that protoc-all supports.
Feel free to reopen if you'd like to discuss further

PR #200, opened 2020-09-09
Fix grpc-1.31.1-r1 build and bump PROTO_VERSION

vtopc:
Currently master is not building:
```
  so:libprotobuf.so.24 (missing):
    required by: grpc-1.31.1-r1[so:libprotobuf.so.24]
  so:libprotoc.so.24 (missing):
    required by: grpc-1.31.1-r1[so:libprotoc.so.24]
```
https://dev.azure.com/namely/protoc-all/_build/results?buildId=122&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=f8ed7bd8-2a7f-56f6-9385-7fc29a8b5b7b&l=339

`grpc-1.31.1-r1` requires `libprotoc.so.24`, which is in the [libprotoc 3.13.0-r0](https://pkgs.alpinelinux.org/package/edge/main/x86_64/libprotoc).

----

As an alternative we could pin grpc to a version, that requires `libprotoc.so.23`, like it was in #190 `1.31.1-r0`?


abe545 (approved):

abe545:
@vtopc unfortunately `apk` drops support for older versions on the edge channel. I don't think alpine is going to work as a base image for us going forward. We switched to pulling packages instead of building them ourselves because of the change to how `grpc-cli` gets built, and `apk` doesn't allow us to have stable builds... I think our best option is to switch to a debian base image.

alex-namely (approved):

PR #202, opened 2020-09-09
pin grpc 1.31.1-r0

vtopc:
Instead of current grpc-1.31.1-r1


vtopc:
```
ERROR: unsatisfiable constraints:
  grpc-1.31.1-r1:
    breaks: world[grpc=1.31.1-r0]
  so:libprotobuf.so.24 (missing):
    required by: grpc-1.31.1-r1[so:libprotobuf.so.24]
  so:libprotoc.so.24 (missing):
    required by: grpc-1.31.1-r1[so:libprotoc.so.24]
  grpc-dev-1.31.1-r1:
    breaks: world[grpc-dev=1.31.1-r0]
```

PR #203, opened 2020-09-10
Do not pin proto version

vtopc:
This is a long term fix for the issue, that was fixed in the #200 

As [Alpine repo drops old packages](https://gitlab.alpinelinux.org/alpine/abuild/-/issues/9996#note_87135) and at some moment grpc-1.31 requirements were switched from libprotoc-3.12 to libprotoc-3.13(as it was updated from `grpc 1.31.1-r0` to `grpc 1.31.1-r1`) it makes no sense to pin on the proto version, let's leave it for autodiscovery and pin grpc version only.


abe545 (approved):

alex-namely (approved):

ido-namely (commented):
🙏

PR #205, opened 2020-09-14
SOR-1932: Pin MINIMUM version of Alpine packages

vtopc:
**Problem:**
Since we have switched to Alpine and [Alpine repo drops old packages](https://gitlab.alpinelinux.org/alpine/abuild/-/issues/9996#note_87135) we have finished with non-rebuildable builds.
As currently, the stable version for grpc is 1.28, and for community/edge repo it's been 1.30 two weeks ago, 1.31 a week ago and now it's only 1.32, which makes pinning the source of issues.

----

**Open questions:**
- Should we have some abstract build number(e.g. `20200914.3` or just git hash) instead of the current "semantic"? As due to this PR one day we might have `namely/protoc-all@1.32_0` with grpc-1.33 inside.

---

As an alternative, we could/should switch to the Debian - #207


Dockerfile
```diff
@@ -105,13 +103,11 @@ RUN set -ex && apk --update --no-cache add \
...
-    protoc~=${proto_version} \
-    protobuf~=${proto_version} \
-    --repository=http://dl-cdn.alpinelinux.org/alpine/edge/main
-
-RUN set -ex && apk --update --no-cache add \
-    grpc~=${grpc_version} \
-    grpc-cli~=${grpc_version} \
+    "protoc>${proto_version}" \
+    "protobuf~=${proto_version}" \
+    "grpc~=${grpc_version}" \
```
alex-namely, line 108: Should this be consistent with prev command and use `>` ?

vtopc: oops

vtopc: fixed, thanks

alex-namely:
@vtopc 
I would probably hard-code (pin) all versions.

alex-namely:
> > @vtopc
> > I would probably hard-code (pin) all versions.
> 
> How? There is no 1.31 in Alpine anymore.

Oh, I see that for protobuf 1.32 is almost gone.

@vtopc , @ido-namely , @abe545 Maybe we should rather ask Microsoft to increase pipeline build timeout or stop supporting grpc_cli?

vtopc:
As an alternative we could pin to grpc-1.28 which is in stable Alpine@v3.12

vtopc:
> @vtopc
> I would probably hard-code (pin) all versions.

We have never been pinning it for real: e.g. line https://github.com/namely/docker-protoc/blob/master/Dockerfile#L60 always downloads the latest version(which is 1.33 now - https://github.com/grpc/grpc-go/blob/master/version.go#L22)

abe545 (approved):

alex-namely (approved):

PR #214, opened 2020-10-21
Fixing json names for swagger when flag is not used

ido-namely:
Latest version of openapiv2 is using json names by default in the swagger files so it needs to be explicityly set to false when flag is not used.
Also:
 - added tests
 - fixed gateway template referencing old github import path of grpc-gateway.


klvmungai (approved):

all/entrypoint.sh
```diff
@@ -60,6 +59,7 @@ DESCR_FILENAME="descriptor_set.pb"
 CSHARP_OPT=""
 SCALA_OPT=""
 OPENAPI_JSON=false
+DEPRICATION_MESSAGE="(deprecated. Please use --with-openapi-json-names)"
```
vtopc, line 62: What about something more specific? e.g.:
```suggestion
JSON_DEPRICATION_MESSAGE="(deprecated. Please use --with-openapi-json-names)"
```

abe545: Also, typo... it should be `DEPRECATION`, not `DEPRICATION`

gwy/templates/main.go.tmpl
```diff
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/gorilla/handlers"
-	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
```
alex-namely, line 17: is it a breaking change?

klvmungai: The gateway generation is actually currently broken without this change with the compilation of the generated gateway failing with 

```
cannot use gwmux (type *"github.com/grpc-ecosystem/grpc-gateway/runtime".ServeMux) as type *"github.com/grpc-ecosystem/grpc-gateway/v2/runtime".ServeMux in argument to namely_reportbuilder_v1.RegisterReportsGrpcServiceHandlerFromEndpoint
```

This is because grpc-gateway master is currently v2 and the module name adds a v2 to the import path.

ido-namely: Good point, I should have clarified. It's actually broken right now because the package name changed. we confirmed the image works with this change.

all/entrypoint.sh
ido-namely: thanks guys, I actually left it there by mistake, decided not to use it since the messages are slightly different when used.

alex-namely (approved):

abe545 (approved):

PR #217, opened 2020-11-19
Add source relative option for protoc-gen-validate

heliuchuan:
Adds a flag to enable the output path for protoc-gen-validate to be source relative.  Otherwise it will use the directory specified in option go_package.


all/entrypoint.sh
```diff
@@ -136,6 +137,10 @@ while test $# -gt 0; do
             fi
             shift
             ;;
+	--validator-source-relative)
```
ido-namely, line 140: Is there any reason for the validator files to be source relative while other generated code is not source relative, 
or can the `go-source-relative` flag be used to set the value for the validator?

ido-namely (commented):
Hi @heliuchuan , thanks for your pr.
I left a comment/question.

all/entrypoint.sh
heliuchuan: Indeed, I did consider using that flag...however I didn't want to check in something that was not backwards compatible.  
For us I don't believe there would be a reason but I can't speak for everyone.

heliuchuan: @ido-namely ping

ido-namely: Hi @heliuchuan, sorry for the delay.
I'll try to test your changes tonight and if everything works well I'll merge this.

ido-namely (commented):
Hi @heliuchuan ,
Sorry for the delay, I tested your changes and I would be happy to approve and cut 1.33_1 
if you don't mind changing [this](https://github.com/namely/docker-protoc/blob/4833be2f2269ad15ff57b86aa861291d93811f4d/variables.sh#L9) value to `1` first.

ido-namely (commented):
Thanks for the contribution.

PR #222, opened 2020-12-11
Allow gateway to use original proto names

nkushpir:
Starting from `v1.33_0` gateway uses new version of [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/tree/v2.0.1) and one of the major changes there is using [camelCase JSON names by default](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2.0.0/docs/_docs/v2-migration.md#we-now-use-the-camelcase-json-names-by-default).
This PR adds a flag to allow using old behavior.


gwy/generate_gateway.sh
```diff
@@ -95,6 +95,10 @@ while test $# -gt 0; do
       fi
       shift
       ;;
+    --use-proto-names)
```
ido-namely, line 98: I think you can remove this since you moved it to the config file template, no?

ido-namely (commented):
can you also bump the build version from 1 to 2 in variables.sh?

ido-namely (approved):

alex-namely (approved):

gwy/templates/main.go.tmpl
```diff
@@ -251,10 +254,27 @@ func SetupMux(ctx context.Context, cfg proxyConfig) *http.ServeMux {
...
+	options := []runtime.ServeMuxOption{
+	    runtime.WithIncomingHeaderMatcher(incomingHeaderMatcher),
+	    runtime.WithOutgoingHeaderMatcher(outgoingHeaderMatcher),
+	}
+
+	if !cfg.useJSONNames {
+		marshaller := &runtime.HTTPBodyMarshaler{
+			Marshaler: &runtime.JSONPb{
+				MarshalOptions: protojson.MarshalOptions{
+					UseProtoNames:   true,
```
nycdotnet, line 266: should this be `!cfg.useJSONNames` ?

nkushpir: It could be, but I feel that setting exact value is more explicit and easier to read than
```UseProtoNames: !cfg.useJSONNames```

nycdotnet: I may not understand Go enough.  Is the config value used somewhere else?

nkushpir: only to check whether we need a custom marshaller : https://github.com/namely/docker-protoc/pull/222/files#diff-bec7b040a15e52d4cd0b8ffc6c7a0a0b9e95fe7cfb9dc4358b56826043db6ff7R262

nycdotnet: I see - nevermind - I should have caught that.  I guess this is fine since we're not supporting the ability to not emit unpopulated / customize discard unknown.  Thanks

nycdotnet (approved):

PR #224, opened 2020-12-14
SOR-2386 - Avoid sending unpopulated fields by default

ido-namely:
Also make discard unknown fields a configurable option.
https://jira.namely.land/browse/SOR-2386


README.md
```diff
@@ -170,9 +170,17 @@ You can configure additional headers to be sent in the HTTP response.
...
 You can also set headers in the your configuration file (e.g `response-headers.some-header-key`)
 
-### Proto names format
+### Marshalling options
+
+#### Proto names format
 By default, gRPC gateway will return proto names as they are in the proto messages. You can change this behavior by setting `gateway.use-json-names: true` and the gateway will use camelCase JSON names.
 
+#### Unpopulated fields
+By default, gRPC gateway will not emit unpopulated fields. You can change this behavior by setting `gateway.emit-unpopulated: true` and the gateway will populate these fields with default values.
```
nycdotnet, line 179: grpc gateway 2 changed the default.  This namely package preserves the default behavior of v1.  Same as with the names.  Don’t know if it’s worth making this distinction.

nycdotnet (approved):
logic looks good to me.  Thanks

README.md
ido-namely: I tried to change the message but I think it came out confusion. I think what's most important is that users understand the default of this docker image instead of the go module it uses.
I changed `gRPC gateway` to `gen-grpc-gateway` to not confuse the two.

nkushpir (approved):

README.md
```diff
@@ -146,32 +148,47 @@ Any headers starting with `Grpc-` will be prefixed with an `X-`, this is because
...
-By default, gRPC gateway will return proto names as they are in the proto messages. You can change this behavior by setting `gateway.use-json-names: true` and the gateway will use camelCase JSON names.
+### Marshalling options
+
+#### Proto names format
+
+By default, `gen-grpc-gateway` will return proto names as they are in the proto messages. You can change this behavior by setting `gateway.use-json-names: true` and the gateway will use camelCase JSON names.
+
+#### Unpopulated fields
+
+By default, `gen-grpc-gateway` will not emit unpopulated fields. You can change this behavior by setting `gateway.emit-unpopulated: true` and the gateway will populate these fields with default values.
```
vtopc, line 187: Hi, is that for generating `omitempty` JSON option?
https://golang.org/pkg/encoding/json/#Marshal

nkushpir: yes, the default behavior has changed in grpc-gateway v2 (https://github.com/grpc-ecosystem/grpc-gateway/blob/v2.0.0/docs/_docs/v2-migration.md#we-now-emit-default-values-for-all-fields) and this PR just sets old behavior as default.

PR #228, opened 2021-01-21
SOR-1963: pin grpc gateway version

kevinsamoei:
The latest version of grpc-gateway changed the way fieldmasks behave, by setting type `FieldMask` to a string instead of an array. See [PR](https://github.com/grpc-ecosystem/grpc-gateway/pull/1820)

This pins the version of grpc-gatway to `v2.0.1` and we can introduce the breaking change gradually.


gwy/templates/go.mod.tmpl
```diff
@@ -0,0 +1,7 @@
+module gateway
+
+go 1.14
+
+require (
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.0.1
```
ido-namely, line 6: I think we can use this file as an opportunity to pin all the current modules that are getting pulled.
I'm guessing that when we build the binary it updates the go mod and go.sum file, which btw, I think we should also add to the template.

ido-namely (commented):
This looks great @kevinsamoei, thank you.
Other than my comment, can you also add a test to verify that we can use fieldmasks?
It would probably mean you'd need to add a new rpc to the test proto.

Also, when we release a new version, we bump  the version [here](https://github.com/namely/docker-protoc/blob/86f2ef38537cd3475c898f436655819f19e9b354/variables.sh#L9).
(Once done we'll also want to figure out how to release a new 1.33 version, but we'll worry about that later)

gwy/templates/go.mod.tmpl
vtopc: > I think we can use this file as an opportunity to pin all the current modules that are getting pulled.

What do you mean? Any examples?

kevinsamoei: yes the `go.sum` file will be created and `go.mod` updated when we build. I agree on pinning the other dependencies as well.
@vtopc the gwy gateway file has other dependencies that could use pinning. see [this](https://github.com/namely/docker-protoc/blob/86f2ef38537cd3475c898f436655819f19e9b354/gwy/templates/main.go.tmpl#L16-L23)

ido-namely (approved):

alex-namely (approved):

PR #229, opened 2021-01-28
Using jsonpb as the default marshaler as per previous grpc gateway

ido-namely:
This reverts the new default marshaling which was introduced when we upgraded to [grpc-ecosystem/grpc-gateway/v2](https://github.com/grpc-ecosystem/grpc-gateway).
Instead of using [protojson](https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson) it uses [jsonpb](https://pkg.go.dev/github.com/golang/protobuf/jsonpb) by default.

The main difference is how fieldmasks are marshaled (string array vs. string; paths as proto names vs. json names)


README.md
```diff
@@ -184,6 +184,10 @@ You can also set headers in the your configuration file (e.g `response-headers.s
 
 ### Marshalling options
 
+#### Setting Marshler version
```
Tochemey, line 187:
```suggestion
#### Setting Marshaller version
``` 😁

Tochemey (approved):

zenyui (approved):

PR #233, opened 2021-03-08
add go_opt support

vincent178:


Xakt:
Hey! We would really like this functionality added, so that we can use it to pass go options for building our protos. Is it possible to have this merged in? Or is there a workaround?

ido-namely:
> Hey! We would really like this functionality added, so that we can use it to pass go options for building our protos. Is it possible to have this merged in? Or is there a workaround?
> 
> Thanks in advance!

Hi @Xakt , thank you for your PR!
I believe we usually expose specific flags which will set go_opt. 
For example `--go-package-map`, `--go-source-relative`.
Is there anything in particular that you are trying to set for your protoc command?

Xakt:
Hi @ido-namely, thanks for your quick reply. To be clear, this isn't my PR, I simply found it 😅 

We would like to set the go_package through the command line, as we want to keep our proto files as clean and language agnostic as possible. I saw here that this is possible by setting it through the go_opts: https://developers.google.com/protocol-buffers/docs/reference/go-generated#package

Is something like this possible?

ido-namely:
> Hi @ido-namely, thanks for your quick reply. To be clear, this isn't my PR, I simply found it 😅
> 
> We would like to set the go_package through the command line, as we want to keep our proto files as clean and language agnostic as possible. I saw here that this is possible by setting it through the go_opts: https://developers.google.com/protocol-buffers/docs/reference/go-generated#package
> 
> Is something like this possible?

Hi @Xakt , thanks for clarifying.
I believe you can use --go-package-map option for that. As you can see here, the value gets passed to `go_opt`

Xakt:
Thanks @ido-namely, but that puts the value in `go_out` and not `go_opt`.

ido-namely:
> Thanks @ido-namely, but that puts the value in `go_out` and not `go_opt`.

Hey @Xakt , I could be wrong but I'm pretty sure it works the same, I believe go_opt is relatively new, Maybe it won't work when we upgrade the go plugin (protoc-gen-go) but for the current version we're using it should be ok.
Could you give it a try and let me know if that doesn't work?

Xakt:
Hey @ido-namely! I've played around a bit, also just running protoc myself. As far as I can tell (but I am in no means an expert 😅), go_opt can not be replaced with go_out. 
To get the go_package using command line arguments, I have to do the following:

`protoc --go_out=gen --go_opt=M<filename>=<pacakge name> <filename>`

I can't find how I would go on merging the two arguments.

Anyway, I think for now we'll make our lives a bit easier and just go with putting `go_package` in the proto files and remove it if and when you add the support.

Thanks for your quick responses and have a nice weekend!

ido-namely:
This PR is stale by now. I could be convinced otherwise but I don't see a reason to add this option.
Specifically mapping go packages is supported via separate flag and in order to modify other options for the plugin
the `namely/protoc` image can be used to access the protoc command directly without the wrapper helper script in `namely/protoc-all`

PR #240, opened 2021-04-19
upgrade grpc version to 1.35

ido-namely:


abe545 (approved):
I'm fine with this, but should we go to 1.36 while we're upgrading?

ido-namely:
> I'm fine with this, but should we go to 1.36 while we're upgrading?

yes, that's happening in my next pr, I didn't want to skip 1.35.

PR #244, opened 2021-05-23
fix: fix swagger path in Dockerfile template

InfernalRabbit:
Returns health of grpc-gateway generator


InfernalRabbit:
I'm sorry, it is ok

Issue #259, opened 2021-08-16
Update grpc_tools_node_protoc_ts version to 5.3.2


masterkidan:
Hello folks, 
Requesting an update to the typescript generation tools to the latest version. A bug was recently fixed https://github.com/agreatfool/grpc_tools_node_protoc_ts/issues/113 in the current version, this issue causes typescript to be broken if some oneOf types are used.

masterkidan:
Some additional context: My team uses the built images from namely for protoc generation, hence need to update the defaults. 
I've opened a PR with the new version updated, please lmk if you have any further questions.

ido-namely:
resolved in #301

PR #260, opened 2021-08-16
fix(grpc_tools_node_ts_version): Bumping to 5.3.2

masterkidan:
This is to addres #259


zaquestion:
@ido-namely this seem reasonable to you?

ido-namely:
Hi @masterkidan , thanks for the contribution and apologies for the major delay.
Any chance you'll be willing to add a test, specifically for the oneof issue which I understand was your main concern?

masterkidan:
Thanks @ido-namely , I am not sure how to add a test here as the bug is not in the code in the docker container ... the underlying tool that is used did add tests for this particular case, can you point an example of how one would add a test here ?

ido-namely:
> Thanks @ido-namely , I am not sure how to add a test here as the bug is not in the code in the docker container ... the underlying tool that is used did add tests for this particular case, can you point an example of how one would add a test here ?

Hi @masterkidan , sure.
So our tests are mostly in this file https://github.com/namely/docker-protoc/blob/master/all/test.sh and the pattern is to call the image using the flags that you are testing/will generate the relevant file.
So you can add a oneof field to the test proto and then once the generated file is created, you can verify that the oneof field was generated as expected.
Feel free to reach out if you need further assistance.

ido-namely:
superseded by https://github.com/namely/docker-protoc/pull/301

Issue #268, opened 2021-10-20
Unable to build locally


abhi18av:
Hi team,

Thanks for the wonderful project!

I wanted to build locally using `make build` , however I keep hitting this error 

```
Step 41/84 : RUN go get -u github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc
 ---> Running in 94a5129c5e1f
unrecognized import path "embed": import path does not begin with hostname
The command '/bin/sh -c go get -u github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc' returned a non-zero code: 1
```

Could you please guide me a bit here?

ido-namely:
Hi @abhi18av,
I appreciate your patience. 
Can you see if you're able to build the image locally now?

abhi18av:
Hi @ido-namely , 

Unfortunately, not yet. I see the following issue before it errors out

```

Step 42/101 : RUN go get -u github.com/ckaznocha/protoc-gen-lint
 ---> Using cache
 ---> 297eceb64d1b
Step 43/101 : RUN set -e &&     GO111MODULE=on go get github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc
 ---> Using cache
 ---> 8e92779e7e26
Step 44/101 : RUN go get -u github.com/micro/micro/cmd/protoc-gen-micro
 ---> Running in 562bfa8c7d57
go: downloading github.com/micro/micro v1.18.0
go get: module github.com/micro/micro@upgrade found (v1.18.0), but does not contain package github.com/micro/micro/cmd/protoc-gen-micro
The command '/bin/sh -c go get -u github.com/micro/micro/cmd/protoc-gen-micro' returned a non-zero code: 1
```

ido-namely:
Hi @abhi18av, interesting I was not able to reproduce, however we released a new version (which included an upgrade to Go and micro module). Mind checking again?
Thanks!

PR #272, opened 2021-11-24
Fix build due to unpinned mismatched version of protoc-gen-doc

ido-namely:
- The build was failing because of an upgrade to protoc-gen-doc that is not compatible.
To solve the problem the protoc-gen-doc module was pinned to the latest version that is compatible.
 - Also:
 -     Added tests for the relevant flag option
 -     Pinned Golang grpc module to match the image's version.


abe545 (approved):

ido-namely:
Fixes #268

PR #273, opened 2021-11-29
Upgrade to Go 1.17

jcburley:
DRAFT for discussion.


variables.sh
```diff
@@ -6,17 +6,18 @@ NAMESPACE=${NAMESPACE:-namely}
...
-GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.39}
-GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.2.1}
-GRPC_GATEWAY_VERSION=${GRPC_GATEWAY_VERSION:-2.0.1}
-UBER_PROTOTOOL_VERSION=${UBER_PROTOTOOL_VERSION:-1.3.0}
-SCALA_PB_VERSION=${SCALA_PB_VERSION:-0.11.0}
+GO_VERSION=${GO_VERSION:-1.17}
+GRPC_VERSION=${GRPC_VERSION:-1.42}
+GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.42}
+GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.3.0}
+GRPC_GATEWAY_VERSION=${GRPC_GATEWAY_VERSION:-2.7.0}
```
ido-namely, line 13: this ia major breaking change. I wouldn't mind upgrading at some point but I'd prefer not to do so in this release.
Is 2.0.1 not supported with go 1.17?

gwy/templates/go.mod.tmpl
```diff
@@ -9,6 +9,6 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
-	google.golang.org/grpc v1.34.0
-	google.golang.org/protobuf v1.24.0
+	google.golang.org/grpc v1.35.0
```
ido-namely, line 12: I would upgrade to the latest grpc version here (1.42.X)

variables.sh
```diff
@@ -6,17 +6,18 @@ NAMESPACE=${NAMESPACE:-namely}
...
+UBER_PROTOTOOL_VERSION=${UBER_PROTOTOOL_VERSION:-1.10.0}
+SCALA_PB_VERSION=${SCALA_PB_VERSION:-0.11.6}
```
ido-namely, line 15: I think that for some of the unrelated changes to go 1.17  like uber,scala, web (unless they are somehow related and I'm missing something), we should upgrade separately.

variables.sh
```diff
@@ -6,17 +6,18 @@ NAMESPACE=${NAMESPACE:-namely}
...
 
-GO_VERSION=${GO_VERSION:-1.14}
-GRPC_VERSION=${GRPC_VERSION:-1.39}
-GRPC_JAVA_VERSION=${GRPC_JAVA_VERSION:-1.39}
-GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.2.1}
-GRPC_GATEWAY_VERSION=${GRPC_GATEWAY_VERSION:-2.0.1}
-UBER_PROTOTOOL_VERSION=${UBER_PROTOTOOL_VERSION:-1.3.0}
-SCALA_PB_VERSION=${SCALA_PB_VERSION:-0.11.0}
+GO_VERSION=${GO_VERSION:-1.17}
+GRPC_VERSION=${GRPC_VERSION:-1.42}
```
ido-namely, line 10: since we're targeting 1.42, the release candidates should be named `v1.42_0-rc.X` instead of `v1.41_0-rc.X` moving forward (v1.41_0-rc.5 is misleading)

variables.sh
jcburley: > this ia major breaking change. I wouldn't mind upgrading at some point but I'd prefer not to do so in this release.
Is 2.0.1 not supported with go 1.17?

I don't recall there being any issues with Go 1.17 on the gRPC Gateway. Is this the `array`->`string` change I noticed? That does seem potentially breaking.

Happy to exclude that.

How about gRPC Web? Leave that be, or upgrade as shown...?

gwy/templates/go.mod.tmpl
jcburley: > I would upgrade to the latest grpc version here (1.42.X)

Hmm, looks like I missed that one!

variables.sh
ido-namely: yeah, the change from array to string would cause unexpected behavior, if it's not necessary I would dedicate a separate version for that upgrade.

I think in general I wouldn't upgrade anything that is not tied to go 1.17 or gRPC 1.42 (I actually expect the gRPC upgrade to also be released separately before we release the go upgrade, but it's fine to have right now because it's the same version that we'll end up using).

jcburley: > 

FYI, just confirmed that 2.0.1 included an update of its Go docker container to 1.15.3.

jcburley:
Am building & testing a new version of this right now.

jcburley:
Local tests passed; pushed changes.

variables.sh
jcburley: Hmm, I just now noticed "web" in this comment; were you referring to GRPC_WEB_VERSION, changed in an earlier comment's chunk? I didn't change that back. Should I?

variables.sh
jcburley: (I've since Closed that `protorepo` PR, btw.)

variables.sh
jcburley: (I'm testing that back-out locally, btw.)

jcburley: (It passed, so I'll go ahead and push that change, as I'm now convinced that's what you were asking me to do.)

Dockerfile
```diff
@@ -102,11 +104,14 @@ RUN go get -u github.com/mwitkow/go-proto-validators/protoc-gen-govalidators
...
 RUN go get -u github.com/coinbase/protoc-gen-rbi
 
-RUN go get github.com/gomatic/renderizer/cmd/renderizer
+RUN go get github.com/gomatic/renderizer/v2/cmd/renderizer
 
 # Origin protoc-gen-go should be installed last, for not been overwritten by any other binaries(see #210)
 RUN go get -u github.com/golang/protobuf/protoc-gen-go
 
+# Need to get this too:
+RUN go get -u github.com/mwitkow/go-proto-validators
```
jtabachnick, line 113: This seems not have been updated in over a year. Is this safe to use?

gwy/templates/go.mod.tmpl
```diff
@@ -9,6 +9,6 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
-	google.golang.org/grpc v1.34.0
-	google.golang.org/protobuf v1.24.0
+	google.golang.org/grpc v1.40.0
```
jtabachnick, line 12: Why `v1.40.0` not `v1.42.0`?

Dockerfile
jcburley: No idea! Not sure whether it's actually needed, in fact. Anyone here know?

ido-namely: Is this instead of what is in line 102?
Shouldn't we specify the version that we need to get?

gwy/templates/go.mod.tmpl
jcburley: I think this is where @ido-namely asked that I back this down from 1.42.0 so as to allow other languages (whose gRPC generators are not up to date yet) to work.

Does that sound right?

ido-namely: not exactly, I think we need our versions to be aligned, if this is a 1.42 image we're releasing, then this package should be 1.42 as well.
If anyone would like to use 1.40, they should use the 1.40 image.
1.42 will be the first Go 1.17 based image.

stephen-lord (approved):

gwy/templates/go.mod.tmpl
jcburley: Updated PR accordingly (after merging your upgrade to 1.41.0 and resolving conflicts).

Dockerfile
jcburley: "Is this safe to use?": Looks like _we_ don't use this at all; it is made available to (external) clients of this repo to optionally generate validators via one or two means. Besides not having been updated in over a year (which, by itself, is not necessarily a bad thing), it hasn't been upgraded, nor perhaps even tested, with any supported versions of Go that I can see.

So I think it neither hurts nor helps _us_ to include (this latest version) or, if we decide, remove it altogether (along with the `all/entrypoints.sh` option that runs it).

"Is this instead of what is in line 102?": Seems that line 102 is no longer necessary with the new line.

"Shouldn't we specify the version that we need to get?": Yes we should, thanks!

jcburley: Now comes the question of what to do about this stuff entirely:

- It's old and doesn't evidently support any supported version of Go
- It's apparently undocumented (unless one considers `all/entrypoint.sh` to be documentation!)
- It's unused internally (here at Namely)
- It's got no tests, e.g. to make sure that an `import` would actually work and/or the `--go-proto-validator` option to that script would not only work, but generate validator code that would actually build with Go 1.17

Do we know whether anyone is actually using it?

Personally I'd rather we just delete it than release it without some decent testing, but the latter would seemingly add substantial time (maybe a day or two?) to pushing this out the door.

Seems like we could (quietly?) drop support for this now and then, _if_ anyone asks, look into what it'd take to bring "modern" (Go-1.17) validation back on a more-leisurely schedule.

@idorekem : What do you think?

jcburley: Update: I just noticed that the `--with-validator` option, which specifies a _different_ validation method, is tested and passes with the updates I've made to the `Dockerfile` to address earlier concerns. (The `all/entrypoints.sh` script rejects specifying both `--with-validator` and `--go-proto-validator`.)

So, it seems (if my assumptions about what's going on here are correct) we would still be supporting validation in the general sense, just not this old way of doing things, if we simply removed it.

ido-namely (approved):
🎉

PR #248, opened 2021-07-22
Ensure GO_SOURCE_RELATIVE option is picked up by gateway plugin

vaijab:

PR #275, opened 2021-12-28
Allow specifying grpc-gateway options

owenthereal:
Allow specifying grpc-gateway options with `--with-gateway OPTS`. For
example, `--with-gateway paths=source_relative`.

### DIFF
--- a/all/entrypoint.sh
+++ b/all/entrypoint.sh
@@ -13,7 +13,7 @@ printUsage() {
     echo " -o DIRECTORY                   The output directory for generated files. Will be automatically created."
     echo " -i includes                    Extra includes"
     echo " --lint CHECKS                  Enable linting protoc-lint (CHECKS are optional - see https://github.com/ckaznocha/protoc-gen-lint#optional-checks)"
-    echo " --with-gateway                 Generate grpc-gateway files (experimental)."
+    echo " --with-gateway OPTS            Generate grpc-gateway files (experimental, OPTS are optional - see https://github.com/grpc-ecosystem/grpc-gateway)."
     echo " --with-docs FORMAT             Generate documentation (FORMAT is optional - see https://github.com/pseudomuto/protoc-gen-doc#invoking-the-plugin)"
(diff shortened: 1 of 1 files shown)


all/entrypoint.sh
```diff
@@ -458,7 +462,7 @@ if [ $GEN_GATEWAY = true ]; then
     mkdir -p ${GATEWAY_DIR}
 
     protoc $PROTO_INCLUDE \
-		--grpc-gateway_out=logtostderr=true:$GATEWAY_DIR ${PROTO_FILES[@]}
+		--grpc-gateway_out=logtostderr=true${GATEWAY_OPTS}:$GATEWAY_DIR ${PROTO_FILES[@]}
```
ido-namely, line 465: Are there other options you are interested in atm other than `source_relative`?

I ask because this tool, for the most part, manage options internally and expose them via specific flags.
For example, it doesn't support adding options on the `--go_out` protoc arg, and instead different flags like `--go-source-relative` are responsible for that.
So for `source_relative`, I would suggest setting that option on `grpc-gateway_out` if `--go-source-relative` is used (There is actually a competing PR for that)

This is not to say there isn't value supporting other options, so please let me know and we can think about it then.

ido-namely (commented):
Hi @owenthereal ,
Thanks for your contribution.

Please see my comments.

all/entrypoint.sh
owenthereal: For now, I only need `source_relative`. But perhaps this would be helpful for other options. I just noticed https://github.com/namely/docker-protoc/pull/248. Seems like a more consistent approach if only `source_relative` is needed. Feel free to close this PR if https://github.com/namely/docker-protoc/pull/248 is more preferable. I'm using my fork right now for this particular feature and will switch once either this PR or #248 is merged

ido-namely: Thank you, I think we should go with #248 approach for now though we're missing some tests. 
If you want to contribute to that PR it would be greatly appreciated :)

ido-namely:
closing for #248

PR #280, opened 2022-01-17
add "native" typescript support

chris-kruining:
I wish to add proper typescript support to the container via the protoc-gen-ts plugin (the code js_out produces is from the previous century).

At first I started to add this module in a custom dockerfile that extends from this one, but I quickly realized that most of the "magic" is done in this shell script.

So in order to help everyone and not just myself I made this PR. I am super new to Docker and also not very experienced with shell scripting, so any help is more than welcome :D

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,7 +123,7 @@ RUN curl -sSL https://github.com/grpc/grpc-web/releases/download/${grpc_web_vers
     
 # Add Typescript support
 RUN npm --global config set user root
-RUN npm --global install protoc-gen-ts
+RUN npm --global install google-protobuf typescript protoc-gen-ts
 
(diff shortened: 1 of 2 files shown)


chris-kruining:
I've just tested the patch, and it seems to work like a charm :D

chris-kruining:
working example:
`docker pull chriskruining/protoc-all:dev`
`docker run -v ``pwd``:/defs chriskruining/protoc-all:dev -o ./gen -d ./src -l typescript`

assuming you are in a directory with your protofiles in a sud-dir called `src`

chris-kruining:
I switched to a custom plugin for my ts generator, so I am closing this as it is no longer relevant to me

Issue #281, opened 2022-01-21
Gitpod does not build docker-protoc


juandemanjon:
This link https://gitpod.io/#https://github.com/namely/docker-protoc opens the repo in Gitpod and the make command fails at step 67/101
```
Step 67/101 : RUN npm config set unsafe-perm true
 ---> Running in e9943670f51a
/bin/sh: 1: npm: not found
The command '/bin/sh -c npm config set unsafe-perm true' returned a non-zero code: 127
```

PR #282, opened 2022-01-21
Fix for #281 (Gitpod does not build docker-protoc)

juandemanjon:
This PR just installs nodejs and npm

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -148,7 +148,7 @@ RUN set -ex && apt-get update && apt-get install -y --no-install-recommends \
 
 # Install latest Node version
 RUN curl -fsSL https://deb.nodesource.com/setup_${node_version}.x | bash -
-RUN apt-get install -y nodejs
+RUN apt-get install -y nodejs npm
 
 # Add TypeScript support
 RUN npm config set unsafe-perm true


abe545:
HI @juandemanjon - thanks for the contribution. Even though the check passed (we gotta look into this), the docker container fails to build with 
```
Step 66/101 : RUN apt-get install -y nodejs npm
 ---> Running in 834f3d71f1db
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 nodejs : Conflicts: npm
 npm : Depends: node-gyp (>= 3.6.2~) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
The command '/bin/sh -c apt-get install -y nodejs npm' returned a non-zero code: 100
```

The other thing to note is that when node 14 is installed, it outputs this helpful message:
```
## Run `sudo apt-get install -y nodejs` to install Node.js 14.x and npm
```

Are you trying to build this with a newer version of node?

juandemanjon:
@abe545 I am not interested in any particular nodejs & npm versions, just being able to build in a Gitpod instance

abe545:
> @abe545 I am not interested in any particular nodejs & npm versions, just being able to build in a Gitpod instance

I don't understand how gitpod works (I've never used it). But if I had to guess, I'd say that this is the line that is suspicious: https://github.com/namely/docker-protoc/blob/master/build.sh#L3 - if the variables aren't loaded then I suspect it is just trying to install the latest versions of the packages

juandemanjon:
I need to investigate more about what is going on, because building the container image should not matter if the host is a gitpod instance.
I am going to close this PR and open a new one if need it

Issue #81, opened 2018-07-27
Any variants for add mypy-protobuf for typing autocomplite?


arturgspb:
Guys, who knows how to just easy add support:
https://github.com/dropbox/mypy-protobuf

I found only golang docker images in docker-protoc. Any ideas?

ido-namely:
fixed in #288 
released in https://github.com/namely/docker-protoc/releases/tag/v1.48_2

PR #288, opened 2022-03-04
Add mypy support

maxjakob:
Enable the `protoc-gen-mypy` plugin through https://github.com/nipunn1313/mypy-protobuf.

Addresses https://github.com/namely/docker-protoc/issues/81.


maxjakob:
Hello @ido-namely ! :wave: 
Could you or somebody from the Namely team trigger the build for this? Would be great to get it in.
Thanks for maintaining this really useful project! :)

build.sh
```diff
@@ -14,6 +14,7 @@ for build in ${BUILDS[@]}; do
         --build-arg go_version=${GO_VERSION} \
         --build-arg uber_prototool_version=${UBER_PROTOTOOL_VERSION} \
         --build-arg scala_pb_version=${SCALA_PB_VERSION} \
+        --build-arg mypy_version=${MYPY_VERSION} \
```
maxjakob, line 17: the code analysis tool complains about missing quotes here, but I'd say it's better to stay consistent with the rest of the file

Dockerfile
```diff
@@ -134,7 +136,8 @@ RUN set -ex && apt-get update && apt-get install -y --no-install-recommends \
     libssl1.1 \
     openjdk-11-jre \
     dos2unix \
-    gawk
+    gawk \
+    python3-pip
```
ido-namely, line 140: I _think_ that you can install this in line 45 and then use it to install `mypy-protobuf`, so that pip does not need to be part of the final image, though I suppose `mypy-protobuf` will then need to be copied over to the new target from wherever it's installed.

Dockerfile
```diff
@@ -115,6 +115,8 @@ FROM debian:$debian_version-slim AS protoc-all
 ARG grpc_version
 ARG grpc_gateway_version
 
+ARG mypy_version
```
ido-namely, line 118: I suspect this would also need to be declared at the top of the file

all/entrypoint.sh
```diff
@@ -432,6 +443,10 @@ if [[ $GEN_RBI == true ]]; then
     GEN_STRING="$GEN_STRING --rbi_out=$OUT_DIR"
 fi
 
+if [[ $GEN_MYPY == true ]]; then
+    GEN_STRING="$GEN_STRING --mypy_out=$OUT_DIR"
```
ido-namely, line 447: should `--mypy_grpc_out` be used here as well?
https://github.com/nipunn1313/mypy-protobuf#grpc

variables.sh
```diff
@@ -14,6 +14,7 @@ GRPC_WEB_VERSION=${GRPC_WEB_VERSION:-1.3.1}
 GRPC_GATEWAY_VERSION=${GRPC_GATEWAY_VERSION:-v2.0.1}
 UBER_PROTOTOOL_VERSION=${UBER_PROTOTOOL_VERSION:-1.10.0}
 SCALA_PB_VERSION=${SCALA_PB_VERSION:-0.11.0}
+MYPY_VERSION=${MYPY_VERSION:-3.3.0}
```
ido-namely, line 17: we leverage [Renovate](https://github.com/namely/docker-protoc/blob/master/.github/renovate.md#how-to-keep-maintaining-renovate-in-this-repo) to manage dependencies.
Would you mind adding the configuration to https://github.com/namely/docker-protoc/blob/master/.github/renovate.json5 ?

I think it should be something like this (I'll try to test it later):
```
  {
    fileMatch: [
      "variables.sh$"
    ],
    matchStrings: [
      "MYPY_VERSION =\\$\\{MYPY_VERSION:-(?<currentValue>.*?)\\}\\n"
    ],
    depNameTemplate: "mypy-protobuf",
    datasourceTemplate: "pypi",
  }
```

ido-namely (commented):
Thank you @maxjakob for the contribution!
I left a few comments but the PR looks really good!

I also suggest to add tests in https://github.com/namely/docker-protoc/blob/master/all/test/all_test.go, should be straightforward.

Might be better if I let the build run after that.

Thanks!

build.sh
ido-namely: I generally agree consistency is better but since I plan to update the project and add those missing quotes everywhere I think it's ok to start adding them here.

Dockerfile
maxjakob: maybe that can work but I have doubts. what we are after is the `protoc-gen-mypy` plugin. when I install that on my system I see that it is a script referring things in my local python environment:
```
$ which protoc-gen-mypy
/home/maxjakob/.pyenv/shims/protoc-gen-mypy
$ cat /home/maxjakob/.pyenv/shims/protoc-gen-mypy
#!/usr/bin/env bash
set -e
[ -n "$PYENV_DEBUG" ] && set -x

program="${0##*/}"

export PYENV_ROOT="/home/maxjakob/.pyenv"
exec "/home/maxjakob/.pyenv/libexec/pyenv" exec "$program" "$@"
```
I'm going to give it a shot and see if I can find all related things to copy from the `build` to the `protoc-all` stage to keep the container size slimg.

maxjakob (commented):
Thanks so much for the feedback @ido-namely! I addressed most of your comments.

Dockerfile
maxjakob: I think I made it work.

.github/renovate.json5
```diff
@@ -180,6 +180,16 @@
       depNameTemplate: "debian",
       datasourceTemplate: "docker",
       versioningTemplate: "debian",
+    },
+    {
+      fileMatch: [
+        "variables.sh$"
+      ],
+      matchStrings: [
+        "MYPY_VERSION =\\$\\{MYPY_VERSION:-(?<currentValue>.*?)\\}\\n"
```
ido-namely, line 189: sorry I added a space by mistake in my example. but the good news is that I tested it and it works well (after removing the space)
```suggestion
        "MYPY_VERSION=\\$\\{MYPY_VERSION:-(?<currentValue>.*?)\\}\\n"
```

ido-namely (commented):
almost there :)

jcburley (approved):

all/test/all_test.go
```diff
@@ -140,6 +140,20 @@ func (s *TestSuite) TestAllCases() {
...
+			lang:              "python",
+			protofileName:     "all/test/test.proto",
+			expectedOutputDir: "gen/pb_python",
+			fileExpectations: []FileExpectation{
+				{fileName: "/__init__.py"},
+				{fileName: "/all/__init__.py"},
+				{fileName: "/all/test/__init__.py"},
+				{fileName: "/all/test/test_pb2.py"},
+				{fileName: "/all/test/test_pb2.pyi"},
+				{fileName: "/all/test/test_pb2_grpc.py"},
```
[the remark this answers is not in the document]
maxjakob: yes, well-spotted! :ok_hand:

ido-namely (approved):

PR #290, opened 2022-03-31
Update to new protoc-gen-go repo, invoke protoc-gen-go-grpc

caseybarker:
Google split `protoc-gen-go` into two parts, adding `protoc-gen-go-grpc`, and moving to a new repository. See this comment for a summary:
https://github.com/golang/protobuf/issues/1070#issuecomment-607293496

The all-in-one `protoc-gen-go` has been deprecated, and the new `protoc-gen-go` requires `protoc-gen-go-grpc` when generating gRPC stubs. The container was already installing `protoc-gen-go-grpc` in Dockerfile:71, but (apparently?) not using it because the old `protoc-gen-go` was still being installed on top in Dockerfile:108.

This PR switches to the newer repo for `protoc-gen-go`, and then changes the `protoc` invocation for Go to use `protoc-gen-go-grpc` instead of the deprecated `plugin` option.


ido-namely:
Thanks for the contribution @caseybarker !
I replaced this PR with some modifications - https://github.com/namely/docker-protoc/pull/325

PR #292, opened 2022-05-05
Bump grpc to 1.43

abe545:
I also got the tests running in the build. Hopefully this will speed up the process with which we can accept external PRs


all/test.sh
```diff
@@ -24,15 +26,21 @@ testGeneration() {
...
     echo "Testing language $lang $expected_output_dir $extra_args"
 
+    mkdir -p "$name"
+    cp -r ./all "./$name"
+    pushd "./$name"
+
     # Test calling a file directly.
     exitCode=0
-    docker run --rm -v=`pwd`:/defs $CONTAINER -f all/test/test.proto -l $lang -i all/test/ $extra_args > /dev/null || exitCode=$?
+    docker run --rm -v=$(pwd):/defs "$CONTAINER" -f all/test/test.proto -l "$lang" -i all/test/ $extra_args > /dev/null || exitCode=$?
```
abe545, line 35: Codacy is yelling about this line, but `$extra_args` can't be quoted - it needs to stay as an array, or the arguments don't get passed to docker correctly

jcburley: `extra_args` doesn't look like it's defined as an array to me. Line 26 should probably be `extra_args=( "$@" )` to make it an array.

Then, instead of `$extra_args`, Codacy might be okay with `"${extra_args[@]}`.

all/test.sh
```diff
@@ -56,6 +64,15 @@ testGeneration() {
         fi
     fi
 
+    if [[ "$lang" == "java" ]]; then
+        if [[ "$extra_args" == *"-o gen/test.jar" ]]; then
```
jcburley, line 68: Not sure this works if `extra_args` is truly an array and the option and its argument are two separate elements in the array.

So maybe you want just a scalar, in which case line 26 should probably be `extra_args="$@"`. (But Codacy might still complain about the quoteless expansion.)

Or, use an array, and change the comparison here to work with that.

jcburley (changes requested):

all/test.sh
jcburley: OR or, convert the array to a string temporarily here and compare that, which might be easiest?

.github/workflows/master.yml
```diff
@@ -36,5 +36,11 @@ jobs:
...
         env:
-          BUILD_VERSION: ${{ steps.sha_short.outputs.SHA_SHORT }}
+          VERSION: ${{ steps.sha_short.outputs.SHA_SHORT }}
         run: make build
+
+      - name: Test
+        if: success()
+        env:
+          CONTAINER: docker.io/namely/protoc-all:${{ steps.sha_short.outputs.SHA_SHORT }}
+        run: make test
```
ido-namely, line 46: we should probably try to run the gwy tests as well - https://github.com/namely/docker-protoc#grpc-gateway-test

gwy/templates/go.mod.tmpl
```diff
@@ -9,6 +9,6 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
-	google.golang.org/grpc v1.42.0
+	google.golang.org/grpc v1.45.2
```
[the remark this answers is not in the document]
ido-namely: I think it should be fine.

ido-namely (commented):
I wonder if we should release 1.43_0 and 1.44_0 before 1.45_0 for the community.

all/test.sh
abe545: I followed the same pattern used throughout this file. I don't think it makes sense to do something differently here than all the other tests

abe545:
> I wonder if we should release 1.43_0 and 1.44_0 before 1.45_0 for the community.

Yes, we probably should stay on top of these and do so (they've also already released 1.46 fwiw)

gwy/templates/go.mod.tmpl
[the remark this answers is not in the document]
abe545: Good thing you got me to run the grpc gateway tests - the point release didn't work, as golang was failing to find it during build. Switching to `v1.45.0` did work though, so I suppose that is how it should be done?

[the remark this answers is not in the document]
ido-namely: yeah, I thought the point release was for the go module, but it looks like `1.45` has only `v1.45.0` - https://pkg.go.dev/google.golang.org/grpc?tab=versions

abe545:
> > I wonder if we should release 1.43_0 and 1.44_0 before 1.45_0 for the community.
> 
> Yes, we probably should stay on top of these and do so (they've also already released 1.46 fwiw)

@ido-namely should I just turn this pr into 1.43_0, so we can have a better PR experience for 1.44_0 and later?

ido-namely:
> > > I wonder if we should release 1.43_0 and 1.44_0 before 1.45_0 for the community.
> > 
> > 
> > Yes, we probably should stay on top of these and do so (they've also already released 1.46 fwiw)
> 
> @ido-namely should I just turn this pr into 1.43_0, so we can have a better PR experience for 1.44_0 and later?

Yes, that's what I was thinking too.

ido-namely (approved):
Thank you for all the work, especially around testing in the CI.

abe545:
> Thank you for all the work, especially around testing in the CI.

I'd love it if we didn't need to do a pr to release a new version, and just base it off the tags... Maybe one day 😄

ido-namely:
> > Thank you for all the work, especially around testing in the CI.
> 
> I'd love it if we didn't need to do a pr to release a new version, and just base it off the tags... Maybe one day 😄

Me too, but on the other hand I do prefer the version to be sourced controlled. 
I think it would be great if we utilize Renovate ([example](https://github.com/ido-namely/docker-protoc/pull/1)) to open these PRs automatically. Now with your addition of the tests to the CI it would make everything much easier.

abe545:
> > > Thank you for all the work, especially around testing in the CI.
> > 
> > 
> > I'd love it if we didn't need to do a pr to release a new version, and just base it off the tags... Maybe one day 😄
> 
> Me too, but on the other hand I do prefer the version to be sourced controlled. I think it would be great if we utilize Renovate ([example](https://github.com/ido-namely/docker-protoc/pull/1)) to open these PRs automatically. Now with your addition of the tests to the CI it would make everything much easier.

Yeah, that seems nice. I wonder if we can enable that for this repo, or if not, if dependabot could be configured to do so

gwy/test.sh
```diff
@@ -1,19 +1,19 @@
...
 
-CONTAINER=$1
 HEADERS_FILE="./.headers"
 SOME_RESP_HEADER="SOME-RESPONSE-HEADER"
 
+pushd "gwy"
+
 # Test building the gateway.
-docker run --rm -v=`pwd`:/defs $CONTAINER -f test/test.proto -i . -s Message
+docker run --rm -v=$(pwd):/defs "$CONTAINER" -f /defs/test/test.proto -i /defs -s Message
```
jcburley, line 16: Probably should be double-quoted, in case there's a space or something in the cwd, and `$PWD` works fine (without spawning a shell) AFAIK, so maybe `"$PWD"`?

gwy/test.sh
```diff
@@ -32,44 +32,54 @@ status=`curl -i -s -o $HEADERS_FILE -w "%{http_code}" localhost:8080/api/message
...
 fi
+echo "[Pass] - Received expected response from gateway when expected payload not passed in the http request body"
 
 kill $!
 
-
-
 # Test building the gateway with unbound methods.
-docker run --rm -v=`pwd`:/defs $CONTAINER -f test/test.proto -i . -s Message --generate-unbound-methods
+docker run --rm -v=$(pwd):/defs "$CONTAINER" -f test/test.proto -i . -s Message --generate-unbound-methods
```
jcburley, line 82: Same here as above (and elsewhere).

all/test.sh
jcburley: Agreed!

all/test.sh
jcburley: Yeah, if you don't need it to be an array, leave it as-is, and just tell Codacy to ignore that issue, is my suggestion, based on what you're saying.

Instead of an array, I'd call it something like "a space-separated string of arguments", FWIW.

jcburley (commented):
Generally, is it expected that error messages will go to `stdout` instead of `stderr` in the pertinent scripts?

Normally I `echo >&2 ...` for errors, but these scripts might be special cases.

all/test.sh
abe545: Fair - once I saw your feedback, I read up on the `$@` and that is indeed what it is - so TIL it isn't an array like in normal programming languages 😄 
If I were an admin in this repo, I'd definitely do that. Unfortunately, I don't have the rights to make codacy ignore it.

gwy/test.sh
abe545: So when I do this locally, that is an empty string in these scripts...

abe545: Oh wait, is it case sensitive? Let me try that

abe545:
> Generally, is it expected that error messages will go to `stdout` instead of `stderr` in the pertinent scripts?
> 
> Normally I `echo >&2 ...` for errors, but these scripts might be special cases.

I can modify it to do that. These are home grown tests that were written by people who might not be that familiar with bash/linux scripting... Like I thought about doing color coding on the pass/fail statements, but I don't really know how to do that, and it seemed like a lot of effort, so I just went with the incremental improvements

all/test.sh
jcburley: I don't have the rights to do that either, but somebody here must!

That being said, it has always struck me as a code smell that wildcard-matching on a variable containing args (which are subject to unwanted globbing, besides the quoting issue that is at least constrained by the callers) is used to "parse" those arguments.

It'd be great to refactor that code to use proper/modern parsing techniques someday.

If `set -f` or whatever turns off the globbing (within that function at least?), that might be a good idea, and perhaps Codacy would notice it and stop complaining?

Also the Codacy explanation suggests an alternative: that of using (moving the command-line composition to) a function. It wasn't obvious to me how that might fix the problem, but perhaps it would?

abe545: Sadly setting `-f` in the shbang didn't convince codacy this is okay... Is it really that important to clear up that warning in codacy? Can't we just ignore it ourselves and merge this anyway, since we know its complaints are meaningless?

jcburley (approved):

PR #210, opened 2020-10-09
SOR-1933 bump latest protoc-gen-micro

vtopc:
https://jira.namely.land/browse/SOR-1933

github.com/micro/protobuf/protoc-gen-go is overwriting protoc-gen-go(regardless if its old github.com/golang/protobuf/protoc-gen-go or new google.golang.org/protobuf/cmd/protoc-gen-go) and version of Feb 5, 2018:
```
$ cd go/bin
$ go get -u github.com/golang/protobuf/protoc-gen-go
$ mv protoc-gen-go protoc-gen-go.origin
$ go get -u github.com/micro/protobuf/protoc-gen-go
$ diff protoc-gen-go protoc-gen-go.origin
Binary files protoc-gen-go and protoc-gen-go.origin differ
```

caused by https://github.com/namely/docker-protoc/pull/140/files

PR #297, opened 2022-05-13
bump protoc-gen-validate to latest version to support validation on proto3 optional

zaquestion:


zaquestion:
Installed `protoc-gen-go` before `protoc-gen-validate` as that's a dependency now. Tho I left the later install just incase the version gets overwritten along the way (inline with #210). I believe the existing tests cover us well enough here. I did go ahead and build all this locally and test generation with protoc-gen-validate and saw the expected results as well. Would deeply appreciate bringing this in as it's blocking a number of other things at my work.

ido-namely (commented):
Thanks @zaquestion for the contribution!
PR LGTM, would you mind adding a test case to confirm the plugin works as expected with a proto file which has the optional attribute?

zaquestion:
Hi @ido-namely need some guidance on how best to move this forward. I added to the `all/test/test.proto` an optional field and executed the tests on it. It causes the `protoc-gen-grpc-gateway` plugin to fail as it's pined to a fairly old version (10 minors behind). Updating that causes some fairly specific tests to break as the generation for some FieldMask seems to change in later versions. Also even if it didn't this PR gains a much larger scope if it's updating multiple deps. 

I believe I can split out into a separate `test_validate.proto` file and write a test against that and not update any other deps, just validating the behavior changes of this PR. Would that be suitable for merge, or would you like some other route?

PR #306, opened 2022-06-03
Consider java lang while forming GEN_STRING with validation enabled

venkatduddu:


all/entrypoint.sh
```diff
@@ -396,6 +396,10 @@ if [[ $GEN_VALIDATOR == true && $GEN_LANG == "gogo" ]]; then
     GEN_STRING="$GEN_STRING --validate_out=lang=gogo${VALIDATOR_SOURCE_RELATIVE}:$OUT_DIR"
 fi
 
+if [[ $GEN_VALIDATOR == true && $GEN_LANG == "java" ]]; then
+    GEN_STRING="$GEN_STRING --validate_out=lang=java${VALIDATOR_SOURCE_RELATIVE}:$OUT_DIR"
```
ido-namely, line 400: VALIDATOR_SOURCE_RELATIVE is a go/gogo flag, so there is no need to set it.

ido-namely (commented):
Thanks for the pr @venkatduddu!
It would be great if you can add a test for this use case.
You can find our tests in https://github.com/namely/docker-protoc/blob/master/all/test.sh

all/entrypoint.sh
venkatduddu: @ido-namely took care of it

venkatduddu:
> Thanks for the pr @venkatduddu! It would be great if you can add a test for this use case. You can find our tests in https://github.com/namely/docker-protoc/blob/master/all/test.sh

@ido-namely tried adding a test for java `--with-validator` flag, however it wasn't generating any validation related java classes/methods. Do I need to modify the test.proto to include some validations?

ido-namely:
> > Thanks for the pr @venkatduddu! It would be great if you can add a test for this use case. You can find our tests in https://github.com/namely/docker-protoc/blob/master/all/test.sh
> 
> @ido-namely tried adding a test for java `--with-validator` flag, however it wasn't generating any validation related java classes/methods. Do I need to modify the test.proto to include some validations?
> 
> image

Hi @venkatduddu , that would be my guess, perhaps in Java it does not generate an additional file if there aren't any annotations.

venkatduddu:
> > > Thanks for the pr @venkatduddu! It would be great if you can add a test for this use case. You can find our tests in https://github.com/namely/docker-protoc/blob/master/all/test.sh
> > 
> > 
> > @ido-namely tried adding a test for java `--with-validator` flag, however it wasn't generating any validation related java classes/methods. Do I need to modify the test.proto to include some validations?
> > image
> 
> Hi @venkatduddu , that would be my guess, perhaps in Java it does not generate an additional file if there aren't any annotations.

Hi @ido-namely you are right, i added validate.proto and it generated the java validator.
Have a look and see how does it look.

all/test.sh
```diff
@@ -187,9 +187,7 @@ testGeneration() {
         fi
     fi
 
-        # Test that we have generated the test.pb.go file.
```
ido-namely, line 190: we'd want to keep this comment

all/test.sh
```diff
@@ -207,6 +205,23 @@ testGeneration() {
         fi
     fi
 
+    if [[ "$extra_args" == *"--with-validator"* ]] && [[ "$lang" == "java" ]]; then
+        # Test that we have generated the Test.java files and TestValidator.java files
+        java_file_count=$(find $expected_output_dir -type f -name "Test.java" | wc -l)
+        if [ $java_file_count -ne 1 ]; then
+            echo >&2 "[Fail] $name"
+            echo >&2 "Test.java files were not generated in $expected_output_dir"
```
ido-namely, line 213:
```suggestion
            echo >&2 "Test.java file was not generated in $expected_output_dir"
```

all/test.sh
```diff
@@ -207,6 +205,23 @@ testGeneration() {
...
+            exit 1
+        fi
+        java_validator_file_count=$(find $expected_output_dir -type f -name "TestValidator.java" | wc -l)
+        if [ $java_validator_file_count -ne 1 ]; then
+            echo >&2 "[Fail] $name"
+            echo >&2 "TestValidator.java files were not generated in $expected_output_dir"
```
ido-namely, line 219:
```suggestion
            echo >&2 "TestValidator.java file was not generated in $expected_output_dir"
```

all/test.sh
```diff
@@ -207,6 +205,23 @@ testGeneration() {
...
+            exit 1
+        fi
+        echo "java with validator"
```
ido-namely, line 222: this can be removed

all/test/test.proto
```diff
@@ -5,6 +5,7 @@ package Messages;
 import "all/test/include.proto";
 import "google/api/annotations.proto";
 import "google/protobuf/field_mask.proto";
+import "all/test/validate.proto";
```
ido-namely, line 8:
```suggestion
import "validate/validate.proto";
```

The file is included in the docker image (in /opt/include/validate/validate.proto) so you don't need to add it to the repo.

all/test.sh
venkatduddu: taken care of it

all/test.sh
venkatduddu: taken care of it

all/test.sh
venkatduddu: taken care of it

all/test/test.proto
venkatduddu: took care of it

ido-namely (approved):

abe545 (approved):

ido-namely:
Now available in `v1.46_1`.
Thank you for your contribution @venkatduddu !

PR #307, opened 2022-06-03
Use latest protogen-validate version that has support for optionals

venkatduddu:
The current version (0.6.1) of proto gen validate used in this doesn't have the optional support.

As part of [0.6.3](https://github.com/envoyproxy/protoc-gen-validate/releases/tag/v0.6.3) protoc-gen-validate started supporting optional.


ido-namely (commented):
Hi @venkatduddu ,
There is another older PR to address this upgrade, I don't mind which one we end up merging but I would prefer if we add some tests.
Please see my comments in the other PR https://github.com/namely/docker-protoc/pull/297

venkatduddu:
> 

i will close this PR.

PR #310, opened 2022-07-05
Rewrite tests in go

ido-namely:
Rewriting the tests in go to make it easier to add new tests and maintain the existing tests.
Also: 
 - Activating go updates in Renovate
 - Removing the need of installing jq in favor of installing go
 - Making the tests run in parallel
 - Fixing some of the existing tests that did not run/validate the existence of generated files properly.

Note: we still have some tests written in bash for the gateway image. This is not addressed in this PR on purpose.


jcburley (commented):
"I just wanna tell you, good luck. We're all counting on you." </meme>

Seriously, this is awesome! I've wanted to see something like this done for quite awhile.

Thanks!!

nycdotnet (approved):
ok.  If you could put a little more in the readme about where to go when a contributor makes a change to add to or update the tests, that would be appreciated.

nycdotnet (approved):
nice!

jcburley (approved):
Nice!!

PR #311, opened 2022-07-07
Update envoy validator and add test for validator optional field

ido-namely:
Replaces https://github.com/namely/docker-protoc/pull/297
Credit to @zaquestion


nycdotnet (approved):
Thanks Ido and original contributor.

Dockerfile
```diff
@@ -96,6 +96,8 @@ RUN set -e && \
 
 RUN go get -u github.com/micro/micro/v3/cmd/protoc-gen-micro
 
+# protoc-gen-go is depended on by protoc-gen-validate, install here and then overwrite later just in case
```
jcburley, line 99: "just in case" is kinda vague; can you be more specific?

jcburley (changes requested):

Dockerfile
ido-namely: updated.
Please also see https://github.com/namely/docker-protoc/pull/311/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R109 for context

nnarh (approved):

PR #312, opened 2022-07-07
Fix Renovate for go

ido-namely:
The go.mod file was not discovered because it's under `test` dir which is ignored by the default preset.


ido-namely:
superseded by https://github.com/namely/docker-protoc/pull/314

PR #314, opened 2022-07-10
Enhance Renovate configuration

ido-namely:
- Manage more dependencies with Renovate
 - Group gateway go package separately (we avoid upgrading this package atm because of a breaking change regarding fieldmasks)
 - Manage gateway go.mod template using Renovate's Go data source
 - Remove go.sum.template (no need for it since the go app is vendored anew in the generated docker file).


nycdotnet (approved):
Would you please make a small update to the readme discussing what Renovate is and how this project uses it?  There is a link to it in the readme already but a few sentences to guide someone like me who may be interested to use the tooling you've set up but isn't quite sure where to get started (some typical commands for example) would be very helpful.

.github/README.md
```diff
@@ -0,0 +1,38 @@
...
+## How does Renovate work?
+
+`Renovate` is able to automatically identify dependencies of different kinds in different files and start opening PRs once a basic configuration file is found in the repo and the app is enabled. `Renovate` has a lot of different configuration options beyond the supported dependencies which are well documented [here](https://docs.renovatebot.com/).
+
+However, in cases such as this repo where the dependencies are managed in proprietary files, additional configuration is required.
+
+## How is Renovate used in this repo?
+
+In this repo, we have a variety of dependencies pinned to specific versions, primarly to create reproducible, testable builds.\
+Most of these dependencies are installed in the built docker images and are referenced in [Dockerfile](../Dockerfile). However, their versions are set separatly in [variables.sh](../variables.sh), which means most PRs opened by `Renovate` will attempt to update this file.
```
nycdotnet, line 18:
```suggestion
Most of these dependencies are installed in the built docker images and are referenced in [Dockerfile](../Dockerfile). However, their versions are set separately in [variables.sh](../variables.sh), which means most PRs opened by `Renovate` will attempt to update this file.
```

.github/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+One of great features of `Renovate` is the ability to group dependency updates into mutual PRs according to different criteria to avoid spamming the repo with too many PRs (1 PR per dependency).
+
+In this repo, we have several groups configured according to specific dependencies and/or dependency types:
+
+*   Major/Minor (main) gRPC updates - these dependencies' updates will be opened in one PR. Merging PRs for this dependency group will be followed by releasing new major version in this repo.
+
+*   All other minor/patch dependency updates - Usually as long as the tests pass for this PR we should be able to merge & release a new patch version in this repo.
+
+*   Go gRPC Gateway module - This dependency is singled out since we cannot update it yet due to a breaking change. once we do update it, we can move this dependency to the same group as the other non gRPC dependencies.
```
nycdotnet, line 33:
```suggestion
*   Go gRPC Gateway module - This dependency is singled out since we cannot update it yet due to a breaking change. Once we do update it, we can move this dependency to the same group as the other non gRPC dependencies.
```

nycdotnet (approved):
Thanks @ido-namely !

.github/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## How does Renovate work?
+
+`Renovate` is able to automatically identify dependencies of different kinds in different files and start opening PRs once a basic configuration file is found in the repo and the app is enabled. `Renovate` has a lot of different configuration options beyond the supported dependencies which are well documented [here](https://docs.renovatebot.com/).
+
+However, in cases such as this repo where the dependencies are managed in proprietary files, additional configuration is required.
+
+## How is Renovate used in this repo?
+
+In this repo, we have a variety of dependencies pinned to specific versions, primarly to create reproducible, testable builds.\
```
jcburley, line 17: Is the trailing backslash needed here? That practice is discouraged here (though I'm not a Markdown expert by any means):

https://www.markdownguide.org/basic-syntax/

README.md
```diff
@@ -295,23 +295,28 @@ the language, the arguments to invoke the image with, the expected generated fil
...
+
 ### Release
 
 Any new gRPC version based release is handled automatically once the relevant [Renovate](https://www.mend.io/free-developer-tools/renovate/) branch is merged to master
 via the CI (Github Action). A patch release should be created manually in Github and the CI workflow will take care of the rest.
 
 #### Contributors
 
-Open a PR and ping one of the Namely employees who have worked on this repo recently.  We will take a look as soon as we can.  
+Open a PR and ping one of the Namely employees who have worked on this repo recently.  We will take a look as soon as we can.\
```
jcburley, line 310: Same question about backslash here.

If it's just being used in lieu of extending the line (i.e. the backslash-newline combo is in lieu of a space), note that the file already has longer lines than this one would be.

jcburley (commented):
A lot of this LGTM, and the Renovate explanation helps give me some context, but I'm not yet enough of an expert to properly review all of this, so will (for now) just Comment.

Please let me know whether you want me to find some time to take a deeper dive in order to reach a go/no-go decision on this.

(Overall this seems like another great improvement to me.)

README.md
ido-namely: honestly the backslashes were added by a linter I used that matches Codacy. Personally I don't feel strongly about it either way.

jcburley: Huh, ok!

klvmungai (approved):

PR #336, opened 2022-09-12
Fix the workflows so that we can auto release a new major version

ido-namely:
The previous solution was not working as expected because one workflow cannot be triggered by an event that itself was triggered by the github-action user.

So instead, the workflow to release new docker images is called explicitly by the workflow that will create a new gh release.
the release to docker can still be triggered however even if the gh release is done manually, which will allow us to keep using it for patch versions.


abe545 (approved):

jcburley (approved):
It seems reasonable to me, but I can't comment on the correctness of the implementation beyond looking for obvious typos, due to not knowing this area very well.

PR #379, opened 2023-08-23
Attempt to fix gateway-gen using invalid reference path to swagger docs

abe545:


abe545:
I'm not going to have time to figure this out, sadly

PR #384, opened 2023-10-23
Upgrade ts_proto dependency

manjabes:
ts-proto has added a few useful options since the last version


DocMarten:
ping @ido-namely

Issue #383, opened 2023-10-22
ts-proto version upgrade


badjilounes:
Hi,

We really like the protoc-all image in our project.
However the ts-proto version is very very old and does not support newest options like `indexOutput`.
Could you please upgrade the `TS_PROTO_VERSION` to the latest version or provide a way to configure the image with specific packages version ?

Thanks,
Lounes

PR #386, opened 2023-11-30
Bump to ts-proto 1.165.0 to fix tsc compile error

jakewins:
The current version has an issue that makes it generate typescript that no longer compiles properly:

```
src/gen/google/protobuf/timestamp.ts:2:23 - error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

2 import * as Long from "long";
```

This error originates from this file in ts-proto:  https://github.com/stephenh/ts-proto/blob/v1.138.0/integration/simple/google/protobuf/timestamp.ts

And is fixed in the 1.165.0 version of the same file: https://github.com/stephenh/ts-proto/blob/v1.165.0/integration/simple/google/protobuf/timestamp.ts


DocMarten:
ping @ido-namely

ramykl:
Bump @ido-namely

ramykl:
Closes #383 #384

